PR #3, opened 2016-05-25
Check if section exists rather than just checking that it's not null.

philmcmahon:
This should fix this error showing up in sentry and reported by central prod: https://app.getsentry.com/the-guardian/workflow/issues/130449721/


clloyd:
👍

PR #7, opened 2016-05-27
Add sensitive, legally sensitive icons to content list.

philmcmahon:
They look like this:
screen shot 2016-11-21 at 14 40 35

I'd like to chat about whether they should be on/off by default. I haven't added the information to the drawer yet.


jennysivapalan:
👍

PR #10, opened 2016-06-26
Correction to allow mouse events on the Restore button

shtukas:
When an item in trashed there is an overlay at z-index 999 which prevents access
to the Restore button. This change reduces the size of the overlay div without
changing the display of the label "Trashed" itself.

### DIFF
--- a/public/components/content-list-drawer/_content-list-drawer.scss
+++ b/public/components/content-list-drawer/_content-list-drawer.scss
@@ -389,7 +389,7 @@ $drawer-breakpoint-small: 1400px;
         position: absolute;
         z-index: 999;
         top: 0;
-        height: 100%;
+        height: 16vw;
         left: 0;
         width: calc(100% - 120px);
         line-height: 100%;


shtukas:
Below are the pictures before and after of the size of the "Trashed" div.

shtukas:
picture 0

picture 1

shtukas:
Tested fine on CODE.

clloyd:
Would a better solution to this to position the overlay relative to Furniture/Management Area, so it's never bigger than that? I'd be worried that relying on a height to not overlap the buttons could cause issues in the future if we redesign that screen again.

clloyd:
👍

akash1810:
I might be missing something obvious, but can't we set the z-index of the buttons to be greater than the watermark? That way it'll work across different screen sizes.

shtukas:
Talking with Akash, we came to the conclusion that the problem is affecting or will be affecting quite few UI elements around the drawer. So we should keep the "Trashed" div horizontal.

PR #11, opened 2016-06-27
Moving the Trashed div to horizontal position to avoid affecting surrounding elements

shtukas:
This change keeps the Trashed div in horizontal position (no rotation), to avoid affecting the surrounding UI elements. The restore button was affected, but actually other elements were affected as well. 

The slight change in the visuals was authorised by Nathan.

### DIFF
--- a/public/components/content-list-drawer/_content-list-drawer.scss
+++ b/public/components/content-list-drawer/_content-list-drawer.scss
@@ -394,7 +394,6 @@ $drawer-breakpoint-small: 1400px;
         width: calc(100% - 120px);
         line-height: 100%;
         font-weight: bold;
-        transform: rotate(-12deg);
         text-align: center;
         color: transparentize($c-red, 0.8);
 


shtukas:
Below, "before" and "after" pictures:

picture 3


picture 5

akash1810:
not a blocker: could make `trashed` centred

👍

ShaunYearStrong:
@shtukas - This may be what you're looking for - https://developer.mozilla.org/en/docs/Web/CSS/pointer-events

shtukas:
@akash1810 
I think it is already.
```
.content-list-drawer.content-list-drawer--trashed .content-list-drawer__full-width {

    position: relative;
    overflow: hidden;

    &:before {
        display: block;
        content: "TRASHED";
        font-size: 16vw;
        position: absolute;
        z-index: 999;
        top: 0;
        height: 16vw;
        left: 0;
        width: calc(100% - 120px);
        line-height: 100%;
        font-weight: bold;
        text-align: center;
        color: transparentize($c-red, 0.8);

    }
```

shtukas:
@ShaunYearStrong
+1 Worth knowing! Thanks. Bookmarked.

akash1810:
its not rendering in centre; that might be because the height isn't 100%? and vertical-align isn't set?

shtukas:
@akash1810 

Oh.... you meant _vertically_ centred. The original design didn't have it so I didn't introduce it.

shtukas:
Tested on CODE.

PR #14, opened 2016-07-06
User Interface for section to tag management in admin area

shtukas:
The (admin) UI for managing the section to tag mapping


shtukas:
The new admin page:

picture 6

shtukas:
Displaying a section. To see existing tags and the tag picker UI

picture 7

shtukas:
The testing was already done for the workflow/datastore update:

This is the initial view  of the writers section of "AU News" on workflow front end:

picture 1

and now the same view a couple of seconds after the tag `technology/apple` has been added against the "AU News" section

picture 2

The section to tag management view was:

picture 3

and the last piece of the puzzle was the tag `technology/apple` present against the "Testing 1825" article:

picture 4

app/controllers/Admin.scala
```diff
@@ -207,5 +212,100 @@ object Admin extends Controller with PanDomainAuthActions {
...
+  val removeSectionTagForm = Form(
+    mapping(
+      "section_id" -> longNumber,
+      "tag_id" -> nonEmptyText
+    )(unAssignTagToSectionFormData.apply)(unAssignTagToSectionFormData.unapply)
+  )
+
+  def sectionsAndTags(selectedSectionIdOption: Option[Long]) = (AuthAction andThen WhiteListAuthFilter).async {
+
+    val selectedSectionOptionFt:Future[Option[Section]] = selectedSectionIdOption match {
```
philmcmahon: Just a thought, but since you're explicitly declaring the type here, and there's no `selectedSection` val, you probably don't need to include the OptionFt bit in the variable name? Just a styling thought...

app/controllers/Admin.scala
```diff
@@ -207,5 +212,100 @@ object Admin extends Controller with PanDomainAuthActions {
...
+        val tagIdsFt: Future[Either[ApiError, List[String]]] = SectionsAPI.getTagsForSectionFt(selectedSectionId).asFuture
+        tagIdsFt.map(_.right.toOption)
+      }
+      case None => Future.successful(None)
+    }
+    for {
+      deskList <- getDesks()
+      sectionListFromDB <- getSortedSections()
+      tagIdsOption <- tagIdsOptionFuture
+      selectedSectionOption <- selectedSectionOptionFt
```
philmcmahon: :trollface:

app/controllers/Admin.scala
```diff
@@ -207,5 +212,100 @@ object Admin extends Controller with PanDomainAuthActions {
...
+    } yield {
+      Logger.info(s"selectedSectionIdOption ${selectedSectionIdOption}")
+      Logger.info(s"selectedSectionOption ${selectedSectionOption}")
+      Logger.info(s"tagIdsOption ${tagIdsOption}")
```
philmcmahon: Maybe just the one log line, or delete them

app/controllers/Admin.scala
```diff
@@ -207,5 +212,100 @@ object Admin extends Controller with PanDomainAuthActions {
...
+  def addSectionTag() = (AuthAction andThen WhiteListAuthFilter) { implicit request =>
+    addSectionTagForm.bindFromRequest.fold(
+      formWithErrors => {
+        BadRequest("failed to execute controllers.admin.addSectionTag()")
+      },
+      data => {
+        // data.sectionId
+        // data.tagId
+        Logger.info(s"controller: addSectionTag: data.sectionId ${data.sectionId}")
+        Logger.info(s"controller: addSectionTag: data.tagId ${data.tagId}")
```
philmcmahon: As above 😄

app/controllers/Admin.scala
```diff
@@ -207,5 +212,100 @@ object Admin extends Controller with PanDomainAuthActions {
...
+  }
+
+  def addSectionTag() = (AuthAction andThen WhiteListAuthFilter) { implicit request =>
+    addSectionTagForm.bindFromRequest.fold(
+      formWithErrors => {
+        BadRequest("failed to execute controllers.admin.addSectionTag()")
+      },
+      data => {
+        // data.sectionId
+        // data.tagId
```
philmcmahon: oops!

app/views/admin/adminLayout.scala.html
```diff
@@ -1,4 +1,4 @@
-@(tab: String)(content: Html)
+@(tab: String, config: play.api.libs.json.JsObject)(content: Html)
```
philmcmahon: Maybe set a default here to avoid passing in `unauthorisedUser`

app/views/admin/desksAndSections.scala.html
```diff
@@ -1,8 +1,8 @@
-@(sections: List[Section], addSectionForm: Form[Section], desks: List[Desk], addDeskForm: Form[Desk], selectedDeskOption: Option[Desk])
+@(config:play.api.libs.json.JsObject ,sections: List[Section], addSectionForm: Form[Section], desks: List[Desk], addDeskForm: Form[Desk], selectedDeskOption: Option[Desk])
```
philmcmahon: as above

app/views/admin/sectionsAndTags.scala.html
```diff
@@ -0,0 +1,126 @@
...
+
+            }
+            case None => {}
+        }
+
+        </div>
+
+        <hr />
+
+        <!--
```
philmcmahon: Maybe delete this

common-lib/src/main/scala/com/gu/workflow/api/SectionsAPI.scala
```diff
@@ -36,4 +46,37 @@ object SectionsAPI {
...
+    Logger.info(s"removeSectionTag: data.tagId ${tagId}")
+    for {
+      res <- ApiResponseFt.Async.Right(deleteRequest(s"sectionTagMapping/${sectionId}/${URLEncoder.encode(tagId, "UTF-8")}"))
+      removeRes <- extractDataResponse[Int](res.json)
+    } yield {
+      removeRes
+    }
+  }
+
+  def getTagsForSectionFt(sectionId: Long): ApiResponseFt[List[String]] = {
```
philmcmahon: As discussed 😄

public/admin.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * 
+ */
+
+console.log("Hello to the admin")
+
+var app1 = angular.module('SectionToTag', []);
```
philmcmahon: what's app1?

public/admin.js
```diff
@@ -0,0 +1,59 @@
...
+app1.controller('tagsPickerAppCtrl', function($scope,$http) {
+    $scope.tag_search_results = []
+    $scope.newSearchFragment = function(){
+        console.log( $scope.searchfragment )
+        // https://content.guardianapis.com/tags?api-key=67ec2e48-12cb-456a-a2ae-d292603e1372&q=apple
```
philmcmahon: I'd remove all console.logs unless they're useful logging that someone who's never looked at the code will understand.

public/admin.js
```diff
@@ -0,0 +1,59 @@
...
+        }).then(function mySuccess(response) {
+            $scope.tag_search_results = [];
+            angular.forEach(response.data.response.results, function(item, key) {
+                this.push(item.id);
+            }, $scope.tag_search_results);
+        }, function myError(response) {
+            console.log(response.statusText)
+        });
+    }
+    $scope.add_section_tag_pairing = function(sectionId,tag){
```
philmcmahon: With javascript camelCase might be more appropriate for function names

app/controllers/Admin.scala
shtukas: As per discussion I am going to leave it as it is, in part because this gives relevance to `selectedSectionOption <- selectedSectionOptionFt`.

app/controllers/Admin.scala
shtukas: Deleted

app/controllers/Admin.scala
shtukas: Done :)

app/controllers/Admin.scala
shtukas: If you meant removing the comments, there're gone.

app/views/admin/adminLayout.scala.html
shtukas: I tried but it doesn't work. I think that it's because `Json.obj()` is not a literal representation of a value.

app/views/admin/desksAndSections.scala.html
shtukas: Same impossibility as above

app/views/admin/sectionsAndTags.scala.html
shtukas: Done.

common-lib/src/main/scala/com/gu/workflow/api/SectionsAPI.scala
shtukas: Removed.

public/admin.js
shtukas: Renamed to `SectionToTagApp`

public/admin.js
shtukas: Removed.

public/admin.js
shtukas: Done.

public/admin.js
```diff
@@ -0,0 +1,55 @@
...
+        $http({
+            method : "GET",
+            url : "https://content.guardianapis.com/tags?api-key="+CONFIG.CAPI_API_KEY+"&q="+encodeURIComponent($scope.searchfragment)
+        }).then(function mySuccess(response) {
+            $scope.tag_search_results = [];
+            angular.forEach(response.data.response.results, function(item, key) {
+                this.push(item.id);
+            }, $scope.tag_search_results);
+        }, function myError(response) {
+            console.log(response.statusText)
```
philmcmahon: Sorry, another pinickity change but maybe just 'error' would be a better function name? And you can use `console.error()`

philmcmahon: Sorry, ignore the previous message! You can just use an anonymous function here and for the rest of the callbacks

shtukas: Done. And thanks for pointing this out. It was the result of indiscriminate copy-pasting from the Angular documentation...

shtukas:
Latest variation has been deployed and tested on CODE.

philmcmahon:
Only other thing would be to use console.error instead of console.log when you're logging an error. Otherwise, 👍 , great work! 

If you'd like to push this out then do go ahead - make sure you check the workflow integration tests and editorial tools integration tests (with the latest version of this branch). It's worth playing with workflow before/after to make sure that the extra step with getContent hasn't slowed it down drastically (this will not be obvious on code as I suspect there's a lot less content there).

app/controllers/Admin.scala
LATaylor-guardian: Stylistically, I often denote Options as such: `maybeMyVariableName`, or in your case maybeFutureSelectedSection.

app/controllers/Admin.scala
```diff
@@ -207,5 +212,87 @@ object Admin extends Controller with PanDomainAuthActions {
...
+      case Some(selectedSectionId) => {
+        for {
+          sections <- getSortedSections()
+        }yield{
+          Some(sections.filter( x => x.id == selectedSectionId ).head)
+        }
+      }
+      case None => Future(None)
+    }
+    val tagIdsOptionFuture: Future[Option[List[String]]] = selectedSectionIdOption match {
```
LATaylor-guardian: There's not much need to return a Future[Option[List[String]]] here, you could instead return a Future[List[String]] by flatmapping depending on how you use it throughout the rest of your code.

shtukas:
@philmcmahon I have made the change from `console.log` to `console.error` to log errors. By the way I have just noticed that `console.error` is not standard in Firefox, but causes no problem with Google Chrome (according to their respective documentations).

app/controllers/Admin.scala
shtukas: Hi Luke. This is a good point. The intended semantics is that if the `selectedSectionId` was set (this is an optional value), then the `tagIds` are also going to be set (and otherwise are optional). So, originally, I made sense to me to have both `Some(thing)` at the same time, or both `None` at the same time. Also, later in the code the HTML code checks whether the list of tags is something or not before displaying a `<h3>Existing Tags</h3>` and then the list of tags. 

That having been said, your suggestion makes me thing that the list of tags should always be a list and the view's details driven by the value of `selectedSectionId`. 

I shall refactor this now :)

shtukas: > the list of tags should always be a list 

... which is going to be empty if either
1. No section was selected ( in which case  `selectedSectionId` is `None`), or
2. Was selected but there was no tag against it. 

Cheers,

shtukas: Done in commit e8dabaa3c5f22780b1761ff8a445b1429093505b

app/controllers/Admin.scala
shtukas: Good idea. I sometime adopt the style of the existing code (meaning the style of the previous devs) for consistency.

shtukas:
Refactoring has been deployed and tested on CODE. Going to merge now.

shtukas:
workflow-integration-tests and editorial-tools-integration-tests have passed.

PR #15, opened 2016-07-07
Remove cloudformation template - moved to editorial-tools-platform repo.

philmcmahon:
In line with new approach of not storing cfn templates ina public repo.

Depends on https://github.com/guardian/editorial-tools-platform/pull/20


shtukas:
I can see the other PR ( on editorial-tools-platform ). Approving both. :+1:

PR #16, opened 2016-07-08
opening route /admin/sections-and-tags to non admin users

shtukas:
This change allows user without admin credentials (the target audience of this change), to access the new 'section to tag' management page, which was implemented in the admin area.

### DIFF
--- a/app/controllers/Admin.scala
+++ b/app/controllers/Admin.scala
@@ -271,7 +271,7 @@ object Admin extends Controller with PanDomainAuthActions {
     }
   }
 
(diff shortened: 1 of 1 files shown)


shtukas:
Deployed to CODE and was tested by CP. Non admin users can access `/admin/sections-and-tags`, but not `/admin/desks-and-sections` and since the routes for the Ajax calls have been opened in commit e934f47aa034577e5bc727fc8de030858546d14e , mapping functionality works fine.

philmcmahon:
👍

PR #17, opened 2016-08-12
Bump to play 2.4 and panda 0.3.0

sihil:
In a bid to adopt panda 0.3.0 we need to move workflow to play 2.4. For good measure we bump to the latest versions of scala and sbt. Ideally we should next move to play 2.5, but this is the minimal move and doesn't need to deal with the DI changes.

Most of the changes are due to modifications to the Play Json API, as they no longer support parsing an option type.

It should be noted that I've dropped a logging test as this seems to be testing Play's crypto classes without testing any of this project's code. If this test is necessary for some reason I'm happy to make the test work, but it seems superfluous.

Thanks to @clloyd for doing a bunch of this work already under https://github.com/guardian/workflow-frontend/tree/cl-update-play


sihil:
Your wish @currysoup, @jharewinton. I've not tried this at all, probably worth getting @philmcmahon or someone else to have a quick look too.

jennysivapalan:
👍

sihil:
We should try to ship this before it is too far behind workflow.

philmcmahon:
Sorry, I just saw this! 👍 from me, will try and get it out this week

philmcmahon:
Just put this on CODE and it exploded. I'll have a look into it today/monday, really ought to get this out soon!

PR #18, opened 2016-09-20
Send through tracking parameter to support tracking tags

clloyd:
Reflecting https://github.com/guardian/flexible-content/pull/2657 this send through tracking tags with the param "tracking" rather than the general "tags"


philmcmahon:
👍

PR #19, opened 2016-10-12
Fix restorer url

philmcmahon:
Restorer has moved to restorer.gutools rather than composer-restorer.gutools. This PR reflects that change


jennysivapalan:
🚫 { 👍 }

PR #20, opened 2016-10-14
Add riffraff recipes to update AMI

philmcmahon:
This change adds two RiffRaff deployment recipes - one which updates the AMI parameter of the cfn template, another which does that and then redeploys the app.


philmcmahon:
Tested on CODE, see the results! https://amiable.gutools.co.uk/instanceAMIs?stack=workflow&stage=&app=workflow-frontend

jennysivapalan:
Nice! We should do this on other projects too. 👍

PR #27, opened 2017-02-16
Broken tagmanager

Reettaphant:
Recover from errors in tagmanager and display an error message to the user if tags are unavailable. @philmcmahon

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -38,7 +38,6 @@ object Application extends Controller with PanDomainAuthActions {
   def getSectionsInDesks(): Future[List[models.api.SectionsInDeskMapping]] = {
     SectionDeskMappingsAPI.getSectionsInDesks().asFuture.map { x =>
       println(s"sections in desks: $x")
-
       x match {
         case Right(mappings) => mappings
         case Left(err) => Logger.error(s"error fetching section desk mappings: $err"); List()
@@ -80,12 +79,13 @@ object Application extends Controller with PanDomainAuthActions {
         "/hyper/tags?limit=100&query=tracking/commissioningdesk/&type=tracking&searchField=path")
         .map(_.getOrElse(List[Tag]())) recoverWith {
           case e: Exception => {
-            Logger.error(s"error in fetching tags: $e")
+            Logger.error(s"error in fetching tags: ${e.getMessage}", e)
             Future(List())
           }
         }
     }
     yield {
+      println("comissioningDesks ", commissioningDesks)
       val user = request.user
 
       val config = Json.obj(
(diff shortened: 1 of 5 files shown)


public/components/stub-modal/stub-modal.html
```diff
@@ -30,9 +30,13 @@ <h4 class="modal-title">{{ modalTitle }}</h4>
             </div>
             <div class="form-group pull-left col-xs-8" ng-if="mode !== 'import'">
                 <label for="stub_cdesk">Commissioning info</label>
-                <div>
+
+                <div ng-if="cdesks.length > 0">
                     <select id="stub_cdesk" name="cdesk" ng-model="stub.commissioningDesks" ng-options="cdesk.id.toString() as cdesk.externalName for cdesk in cdesks"></select>
                 </div>
+                <div ng-if="cdesks.length === 0">
+                    <p class="help-block">Selecting comissioning info is temporarily unavailable</p>
```
philmcmahon: typo - commissioning

app/controllers/Application.scala
```diff
@@ -75,7 +75,14 @@ object Application extends Controller with PanDomainAuthActions {
       sections <-  getSortedSections
       desks <- getSortedDesks
       sectionsInDesks <- getSectionsInDesks
-      commissioningDesks <- TagService.getTags(Config.tagManagerUrl+ "/hyper/tags?limit=100&query=tracking/commissioningdesk/&type=tracking&searchField=path").map(_.getOrElse(List[Tag]()))
+      commissioningDesks <- TagService.getTags(Config.tagManagerUrl+
```
philmcmahon: It was me that made this a bit messy in the first place, but would it be possible to move the logic here into TagService so that the for comprehension is a bit more readable?

app/controllers/Application.scala
```diff
@@ -75,7 +75,14 @@ object Application extends Controller with PanDomainAuthActions {
...
+        "/hyper/tags?limit=100&query=tracking/commissioningdesk/&type=tracking&searchField=path")
+        .map(_.getOrElse(List[Tag]())) recoverWith {
+          case e: Exception => {
+            Logger.error(s"error in fetching tags: $e")
```
philmcmahon: When logging exceptions, I think it's better to do Logger.error("error fetching tags: ${e.getMessage}", e) - that way the entire exception gets passed to the logger and can be handled nicely, rather than the full string representation with stack tract etc. being printed in the message.

Reettaphant:
@philmcmahon I've addressed your comments now.

philmcmahon:
Looks good! 👍

PR #28, opened 2017-03-23
Use CAPI to render the contents of the item drawer.

philmcmahon:
As part of the 'get atoms in workflow' project, we want to simplify the amount of composer-specific stuff in workflow. This is stage one of that - fetching all fields that are not displayed in the content list on the client side at the point when the content list drawer is opened.

Tested on CODE. Best to deploy this early on in the day as it's quite a major change (major change) to a pretty fundamental bit of workflow.


philmcmahon:
I worked on this with @marialivia16 so going to go ahead and merge it.

PR #33, opened 2017-04-19
Update Node and Node-Sass to latest

gtrufitt:
Simple change to use the latest version of Node and Node-Sass and the related updates to ensure compilation.

I've updated the circle-ci build to use [Ubuntu Trusty](https://circleci.com/docs/1.0/build-image-trusty/#nodejs) which is the image that has Node 6.1.0 on. Seem to build fine on the branch.

image


.nvmrc
```diff
@@ -0,0 +1 @@
+6.1.0
```
akash1810: any particular reason for `6.1.0` and not a later version?

gtrufitt: Because `6.1.0` is the latest pre-installed version on the circle CI images

public/components/content-list-item/_content-list-item.scss
```diff
@@ -380,7 +380,7 @@
     .content-list-item__field--working-title,
     .content-list-item__field--headline {
         @include smallScreen {
-            @extend %fs-data-1;
+            @include font-size(11, 14); // Can't extend placeholders (%fs-data-1) into @media so including the font-size mixin directly
```
shaundillon: I reaaaaaallly hate `@extend` anyway. Not one for now but we should look at getting rid of them.

gtrufitt: Yeah, me too 👍

PR #35, opened 2017-04-21
[WIP] - Upgrade to Angular 1.6.x

shaundillon:
Keeping us up to date.

I've read through the [migration guide](https://docs.angularjs.org/guide/migration) and it's been a pretty direct upgrade path. I'll need to do further testing when Workflow code is more stable.

Next: Dependencies audit and possible datetimepicker refactor! (It's rather excessive that we need bootstrap, angular-bootstrap and a datetimepicker just for one element).


shaundillon:
Closing as this is being introduced seperately

PR #39, opened 2017-04-25
Switch out JSPM for webpack

shaundillon:
Work in progress, looking at normalising the build process in tools and speeding up serving JS.

#To do

- [x] import globals (Presence) correctly

- [x] Move CS build process out of SBT


circle.yml
```diff
@@ -17,10 +17,9 @@ dependencies:
     - "project/target/streams"
   pre:
     - cd $(dirname "$0");
-    - rm -rf node_modules/ public/jspm_packages/
+    - rm -rf node_modules/
     - npm install
```
gtrufitt: Worth switching to using [yarn on circleci](https://circleci.com/docs/1.0/yarn/) as we recommend it locally?

setup.sh
```diff
@@ -1,27 +1,23 @@
...
 
+printf "\n\rSetting up client side dependancies... \n\r\n\r"
+printf "\n\rInstalling NPM packages via yarn... \n\r\n\r"
 
-echo "Installing jspm client-side packages"
-./node_modules/.bin/jspm install -y
+yarn
 
-JSPM_EXIT=$?
+printf "\n\Compiling Javascript... \n\r\n\r"
```
gtrufitt: You're missing the `r` between `\n\rCompiling` here I think.

circle.yml
shaundillon: Workflow uses teamcity now so I can remove this file all together I think. @philmcmahon ?

README.md
```diff
@@ -6,13 +6,13 @@ The Angular frontend and associated Scala API for Workflow.
 In order to run, workflow-frontend needs to talk to the CODE workflow datastore. It does this via an SSH tunnel to a 
 workflow-frontend CODE instance.
 
-- Make sure that you are running the right version of nodejs, should be v0.10.45. 
-- Run the install script `./setup.sh` 
+- Make sure that you are running the right version of nodejs, (tested and working with v6.1.0). 
+- Run the install script `./scripts/setup.sh`
```
shaundillon: I've tend to agree but as this is used on teamcity too, it'd break on there. Maybe an improvement in the future would be to push all this to teamcity build steps and use this file solely for local dev.

gtrufitt: Yeah, possibly. No worries then!

PR #43, opened 2017-05-10
Fix presence error due to array containing duplicate values.

philmcmahon:
This is something leftover from the underscore to lodash migration - the syntax for specifying an iterator is different in lodash.


philmcmahon:
Relevant docs: https://lodash.com/docs/#uniqBy

PR #46, opened 2017-05-11
Match the new routes from workflow

marialivia16:
This matches the changes to the routes and the common-lib from workflow.

See https://github.com/guardian/workflow/pull/996


philmcmahon:
4trr2ulbwiia8
👍

PR #47, opened 2017-05-12
Use riffraff.yaml

philmcmahon:
The future is now.


conf/riff-raff.yaml
```diff
@@ -0,0 +1,17 @@
+stacks: [workflow]
+regions: [eu-west-1]
+deployments:
+  workflow-frontend:
+    type: autoscaling
+    parameters:
+      bucket: workflow-dist
```
sihil: you need a `dependsOn: workflow-frontend-ami-update` here otherwise you'll have a race condition on the AMI CFN update and the autoscaling group which has been known to cause entertaining effects (e.g. the ASG increases the number in the ASG and the CFN update knocks it back down again).

philmcmahon: Thanks!

PR #49, opened 2017-05-16
Adds editorId to stub

marialivia16:


common-lib/src/main/scala/models/Stub.scala
```diff
@@ -22,6 +22,7 @@ case class Stub(id: Option[Long] = None,
                lastModified: DateTime = DateTime.now(),
                trashed: Boolean = false,
                commissioningDesks: Option[String] = None,
+               editor_id: Option[String] = None,
```
philmcmahon: should this be editorId?

philmcmahon:
woosh

PR #56, opened 2017-05-17
fix [$rootScope:inprog] $digest already in progress

shaundillon:
as per http://www.pro-tekconsulting.com/blog/solution-to-error-digest-already-in-progress-in-angularjs-2/

Checks if an `$apply` or `$digest` is in progress before running, will cut out a lot of noise from Sentry.


philmcmahon:
👍

PR #61, opened 2017-05-23
Create media atoms from workflow

marialivia16:
This feature is behind the `support-atoms` switch. Go to `feature/support-atoms/on` to activate it.

**TODO**
- [ ] Add an atom icon
- [ ] Make sure we are never setting the status to `null`, we should always have a value for status as in the future we will make the column required. This means that the status of a `news list` has to be set to be `Stub`.
- [ ] Hide atoms behind a feature switch

### DIFF
--- a/public/components/stub-modal/stub-modal.js
+++ b/public/components/stub-modal/stub-modal.js
@@ -142,12 +142,13 @@ function StubModalInstanceCtrl($rootScope, $scope, $modalInstance, $window, conf
         const stub = $scope.stub;
         let promise;
 
-        if (!addToComposer && !($scope.mode === "import")) {
+        if ((!addToComposer || !addToAtomEditor) && !($scope.mode === "import")) {
             delete stub.status;
         }
 
         if ($scope.contentName === 'Atom') {
             stub['contentType'] = $scope.stub.contentType.toLowerCase();
+            stub['status'] = stub.status === undefined ? 'Stub' : stub.status;
             if (addToAtomEditor) {
                 promise = wfContentService.createInMediaAtomMaker(stub);
             } else if (stub.id) {
--- a/public/components/stub-modal/stub-modal.js
+++ b/public/components/stub-modal/stub-modal.js
@@ -255,6 +255,8 @@ wfStubModal.run([
     'config',
     function ($window, $rootScope, $modal, $log, wfContentService, wfFiltersService, wfProdOfficeService, wfPreferencesService, wfLocationService, sections, config) {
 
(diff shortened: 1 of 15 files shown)


public/components/stub-modal/stub-modal.js
```diff
@@ -134,26 +138,34 @@ function StubModalInstanceCtrl($rootScope,$scope, $modalInstance, $window, confi
...
-            delete stub.status;
-        }
-
-        if (addToComposer) {
-            promise = wfContentService.createInComposer(stub);
-        } else if (stub.id) {
-            promise = wfContentService.updateStub(stub);
+    $scope.ok = function (addToComposer, addToAtomEditor) {
+        const stub = $scope.stub;
+        let promise;
```
akash1810: minor: this could be its own function to avoid the `let`, e.g:

```js
function foo() {
  if ($scope.contentName === 'Atom') {
    return wfContentService.createInMediaAtomMaker(stub);
  } else {
    return wfContentService.createStub(stub);
  }
}
```

public/lib/content-service.js
```diff
@@ -1,24 +1,39 @@
...
+                            "article": "Article",
+                            "liveblog": "Live blog",
+                            "gallery": "Gallery",
+                            "interactive": "Interactive",
+                            "picture": "Picture"
+                        };
+
+                        if (isActive) basicTypes['atom'] = 'Atom';
+
+                        return basicTypes;
```
akash1810: could use `Object.assign` to avoid the mutation:

```js
const basicTypes = {'article': 'Article', ...}
return isActive
  ? Object.assign({}, basicTypes, {'atom': 'Atom'})
  :  basicTypes;
```

public/components/stub-modal/stub-modal.js
```diff
@@ -228,6 +255,8 @@ wfStubModal.run([
     'config',
     function ($window, $rootScope, $modal, $log, wfContentService, wfFiltersService, wfProdOfficeService, wfPreferencesService, wfLocationService, sections, config) {
 
+        var defaultAtomType = "media"
```
akash1810: can be a `const`

public/components/stub-modal/stub-modal.js
```diff
@@ -248,7 +277,7 @@ wfStubModal.run([
             function createStubData (contentType, sectionName) {
 
                 return {
-                    contentType: contentType,
+                    contentType: contentType == "atom" ? defaultAtomType : contentType,
```
akash1810: `===` for a stricter comparison

PR #62, opened 2017-06-02
put back route which was (accidentally?) removed

paulmr:
removal of this endpoint breaks composer -> workflow integration.


shaundillon:
LGTM!

PR #63, opened 2017-06-06
When switch is on set supportAtoms parameter to true

marialivia16:
This sends the value of the `support-atoms` feature switch to the backend. When it's `off` we should exclude atoms when querying.

https://github.com/guardian/workflow/pull/1008

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -48,19 +48,16 @@ object Api extends Controller with PanDomainAuthActions {
   }
 
   // can be hidden behind multiple auth endpoints
-  private def getContentBlock[R] = { implicit req: R =>
+  private def getContentBlock[R <: Request[_]] = { implicit req: R =>
     val qs: Map[String, Seq[String]] = req match {
-      case r: UserRequest[_] =>
-        val query = r.queryString + ("email" -> Seq(r.user.email))
-
-        val supportAtoms = r.cookies.get("support-atoms").fold(false)(cookie => cookie.value == "1")
-        if(supportAtoms) query + ("supportAtoms" -> Seq(supportAtoms.toString)) else query
-
-      case r: Request[_] =>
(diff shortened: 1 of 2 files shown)


app/controllers/Api.scala
```diff
@@ -49,9 +49,16 @@ object Api extends Controller with PanDomainAuthActions {
...
+      case r: UserRequest[_] =>
+        val query = r.queryString + ("email" -> Seq(r.user.email))
+
+        val supportAtoms = r.cookies.get("support-atoms").fold(false)(cookie => cookie.value == "1")
+        if(supportAtoms) query + ("supportAtoms" -> Seq(supportAtoms.toString)) else query
+
+      case r: Request[_] =>
+        val supportAtoms = r.cookies.get("support-atoms").fold(false)(cookie => cookie.value == "1")
+        if(supportAtoms) r.queryString + ("supportAtoms" -> Seq(supportAtoms.toString)) else r.queryString
     }
```
paulmr: No really a big deal, but you could consider removing the duplicate code here by doing something like deciding the additions in a advance and then just adding them on the end:

```scala
val supportAtoms =
  if(r.cookies.get("support-atoms").fold(false)(cookie => cookie.value == "1"))
    List("supportAtoms" -> Seq(supportAtoms.toString))
  else
    Nil
// ...
qs ++ supportAtoms
```

marialivia16: The cookies logic can't be taken out of the pattern match because the request is of type `R` 😢 We can probably think of something to remove the `if`, but since this code is just temporary it might not worth it.

paulmr: Even with:

```scala
 private def getContentBlock[R <: Request[_]] = { implicit req: R =>
```

? Still, I take your point that it's only temporary, and it is just a minor comment!

PR #67, opened 2017-06-09
Make status field mandatory

marialivia16:
`workflow` work: https://github.com/guardian/workflow/pull/1010

### DIFF
--- a/common-lib/src/main/scala/com/gu/workflow/api/ApiUtils.scala
+++ b/common-lib/src/main/scala/com/gu/workflow/api/ApiUtils.scala
@@ -42,7 +42,7 @@ object ApiUtils {
   // this function is needed to convert the json into a format that datastore understands.
   def flatStubJsonToStubJson(jsValue: JsValue): ApiResponseFt[JsValue] = {
     jsValue.validate[Stub](flatJsonReads).fold(e => {
-      ApiResponseFt.Left(ApiError("Json conversion failed", s"Failed to convert flat stub into stub with externalData level for datastore with error: $e. Json: $jsValue", 400, "badrequest"))
+      ApiResponseFt.Left(ApiError("Json conversion failed", s"Failed to convert flat stub into stub with externalData level for datastore with error: $e", 400, "badrequest"))
     }, s => ApiResponseFt.Right(Json.toJson(s)(stubWrites)))
   }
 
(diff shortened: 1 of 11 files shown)


common-lib/src/main/scala/com/gu/workflow/api/ApiUtils.scala
```diff
@@ -68,8 +68,8 @@ object ApiUtils {
       case JsSuccess(a, _) => ApiResponseFt.Right(a)
       case error@JsError(_) =>
         val errMsg = errorMsgs(error)
-        Logger.error(s"JsonParseError failed to parse the json. Error(s): $errMsg 400 badrequest")
-        ApiResponseFt.Left(ApiError("JsonParseError", s"failed to parse the json. Error(s): $errMsg", 400, "badrequest"))
+        Logger.error(s"JsonParseError failed to parse the json. Error(s): $errMsg 400 badrequest. Json: $jsValue")
```
susiecoleman: I don't know what the json response would be but is there going to be anything sensitive about our users that we wouldn't want to log? Though we clear out logs relatively quickly so I doubt it would be an issue

common-lib/src/main/scala/models/Status.scala
```diff
@@ -19,11 +19,12 @@ object Status {
   }
   def Stub = Status("Stub")
   def Writers = Status("Writers")
-  def Subs = Status("Subs")
+  def Desk = Status("Desk")
```
susiecoleman: Is there any particular reason why is there a mix between using `Status.Writers` and `Status("Writers")`?

public/lib/content-service.js
```diff
@@ -146,7 +145,7 @@ angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService
                  */
                 updateField(contentItem, field, data) {
 
-                    if (field === 'status' && !contentItem.composerId) {
+                    if (field === 'status' && contentItem.status === 'Stub') {
```
susiecoleman: So stubs don't have a composer id?

shaundillon: @susiecoleman Nope, Stubs are created when you start content in "News List" this means it hasn't got a related composer article yet.

common-lib/src/main/scala/com/gu/workflow/api/ApiUtils.scala
marialivia16: I'll remove the json from the log, I was debugging something 👍

marialivia16:
@susiecoleman Duplication is expected because `workflow` and `workflow-frontend` share a common module called `common-lib`. In the future I'd like to take that bit out in a separate repo to avoid repetition.

public/lib/content-service.js
susiecoleman: Thanks

susiecoleman:
Thanks for the extra workflow info. Looks good to me 👍

PR #73, opened 2017-06-19
Media atom details

shaundillon:
Adds improved details to media atoms and initial hooks for accordion

## Before
picture 216

### With Preview
picture 217

### Without Preview
picture 218


public/lib/trusted-url.js
```diff
@@ -0,0 +1,8 @@
+import angular from 'angular';
+
+angular.module('wfTrustedUrl', [])
```
shaundillon: this is because Angular (rightfully) doesn't trust injected external URLs in `ng-src`

gtrufitt:
Minor thing but I'm wondering if this headline field is going to end up being too tight? 
image

public/components/content-list-drawer/content-list-drawer.html
```diff
@@ -1,10 +1,10 @@
 <tr class="content-list-drawer content-list-drawer--hidden" ng-class="{'content-list-drawer--trashed': contentItem.item.trashed}">
     <td colspan="100" class="content-list-drawer__full-width">
 
-        <div class="drawer">
-            <div class="drawer__section drawer__section--wide">
-                <div class="drawer__header">
-                    <i class="drawer__icon" wf-icon="furniture"/>
+        <div class="drawer" wf-content-list-drawer-accordion wf-open-section="furniture">
```
gtrufitt: Is `wf-content-list-drawer-accordion` a `attr` or should it be a class?

public/components/content-list-drawer/content-list-drawer.html
```diff
@@ -160,31 +160,113 @@
...
+                                    <a href="{{ usage.composerUrl }}" target="_blank"><i class="drawer__icon" wf-icon="composer"></i></a>
+                                    <a href="{{ usage.viewerUrl }}" target="_blank"><i class="drawer__icon" wf-icon="preview"></i></a>
+                                    <a href="{{ usage.webUrl }}" target="_blank"><i class="drawer__icon" wf-icon="live-site"></i></a>
                                 </div>
                             </div>
                         </li>
                     </ul>
+
+                    <ul class="drawer__column">
+                        <li class="drawer__item">
```
gtrufitt: Would it be appropriate to use `ngRepeat` in these drawers? Feels like it might be easier to modify and reason with when we add more content types if this was built up as an array of objects and then looped over?

shaundillon:
@gtrufitt - That's what the expanding drawers (which I'm working on now) are for.

gtrufitt:
@shaundillon Ah course

public/components/content-list-drawer/content-list-drawer.html
```diff
@@ -244,16 +326,17 @@
                             </div>
                         </li>
                         <li class="drawer__item">
-                            <span class="drawer__item-title">In use by</span>
-                            <wf-presence-indicators presence-id="contentItem.composerId" in-drawer="true"></wf-presence-indicators>
+                            <p class="drawer__item-title">In use by</p>
+                            <wf-presence-indicators presence-id="contentItem.composerId" in-drawer="true" ng-if="contentItem.composerId"></wf-presence-indicators>
+                            <wf-presence-indicators presence-id="contentItem.item.editorId" in-drawer="true" ng-if="contentItem.contentType === 'media'"></wf-presence-indicators>
```
gtrufitt: Should we identify the difference between composer and media in the labels?

public/components/content-list-item/content-list-item.js
```diff
@@ -62,6 +62,9 @@ function wfContentItemParser(config, statusLabels, sections) {
                 this.live = LIVE_PATH + item.path;
                 this.ophan = OPHAN_PATH + item.path;
             }
+            if (item.contentType === 'media') {
+                this.editor = `${config.mediaAtomMakerViewAtom}${item.editorId}`
```
gtrufitt: Just checking - are you missing a slash between these?

public/components/content-list-item/templates/links.html
```diff
@@ -1,27 +1,37 @@
...
+<td class="content-list-item__field--preview"  ng-if="contentItem.contentType !== 'media'">
     <a href="{{ contentItem.links.preview }}" title="Preview (opens in new window)" target="_blank" ng-if="contentItem.links.preview" ng-click="$event.stopPropagation()">
         <i class="content-list-item__icon--preview" wf-icon="preview"></i>
     </a>
                 <span ng-if="!contentItem.links.preview" title="Preview (unavailable)" class="content-list-item__icon--inactive">
                     <i class="content-list-item__icon--preview" wf-icon="preview" wf-icon-active="false"></i>
                 </span>
 </td>
-<td class="content-list-item__field--live">
+<td class="content-list-item__field--live"  ng-if="contentItem.contentType !== 'media'">
```
gtrufitt: This can probably be done later, but is this check  `contentItem.contentType !== 'media'` going to be `!== 'atomofanykind` on these at some point?

public/lib/atom-service.js
```diff
@@ -0,0 +1,31 @@
...
+
+    return Object.assign({}, atom, mediaAtomFields);
+
+
+    function getCurrentAsset() {
+      return atom.assets.filter(asset => asset.version === atom.activeVersion)[0];
+    }
+
+    function getUrl() {
+        return `${config.mediaAtomMakerViewAtom}${atom.id}`;
```
gtrufitt: Ah you've done it here too, so I assume the lack of `/` is intentional :)

public/components/content-list-drawer/content-list-drawer.html
shaundillon: It was my original idea for an arribute based directive for the toggling. I'm gonna pull it out now as I moved this up to the drawer controller in #77

public/components/content-list-drawer/content-list-drawer.html
shaundillon: Hmm, possibly, but as each row is a new list it'd be difficult. We'd either have to create arrays of content for each column, which is just shifting the def from the view to the controller, or have some smart logic for starting/ending each list.

public/components/content-list-drawer/content-list-drawer.html
shaundillon: I think the icon on the row and the change in link icons should cover this.

public/components/content-list-item/content-list-item.js
shaundillon: Nope, the snappily named `mediaAtomMakerViewAtom` config item already comes with the slash.

public/components/content-list-item/templates/links.html
shaundillon: Yep, or, ideally we'd have something that denotes *all* atoms and then only use the types once we're sure.

public/lib/atom-service.js
shaundillon: 👍

public/components/content-list-drawer/content-list-drawer.html
gtrufitt: Cool

public/components/content-list-drawer/content-list-drawer.html
gtrufitt: Fair enough, lets see whether we get loads of logic in here once we've got more atoms that would make sense in a model then, but leave it for now 👍

public/components/content-list-drawer/content-list-drawer.html
```diff
@@ -160,31 +160,113 @@
...
+                            </span>
+                        </li>
+
+                    </ul>
+
+                    <ul class="drawer__column drawer__column--wide" ng-if="capiData.youtubeUrl">
+                        <li class="drawer__item">
+                            <p class="drawer__item-title">Video preview</p>
+                            <iframe
+                                type="text/html"
```
akash1810: can an iframe be any other type? 🤔

public/components/content-list-drawer/content-list-drawer.html
```diff
@@ -160,31 +160,113 @@
...
+                                className="drawer__item"
+                                ng-src="{{capiData.youtubeUrl | wfTrustedUrl}}"
```
akash1810: this won't work for self hosted videos, but lets ship this first and go again as we're not self host yet

public/components/content-list-drawer/content-list-drawer.js
```diff
@@ -29,6 +29,19 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
         return config.composerRestorerUrl + '/' + composerId + '/versions';
     }
 
+    function isAtom(contentType) {
+        const allAtomTypes = [
```
akash1810: should this list come from the server?

public/lib/atom-service.js
```diff
@@ -0,0 +1,31 @@
...
+
+function wfAtomService(config) {
+  function parseMediaAtom(atom) {
+    const currentAsset = getCurrentAsset();
+    const editorUrl = getUrl();
+    const friendlyExpiryDate = moment(atom.metadata.expiryDate).format('dddd, MMMM Do YYYY');
+    const mediaAtomFields = {
+      friendlyExpiryDate: friendlyExpiryDate,
+      editorUrl: editorUrl,
+      youtubeUrl: currentAsset && `https://www.youtube.com/embed/${currentAsset.id}?showinfo=0`
```
akash1810: as part of PfP the player looks the same regardless of what query strings are given, so we can remove `showinfo`

see https://github.com/guardian/media-atom-maker/pull/409 and https://github.com/guardian/frontend/pull/16240

public/components/content-list-drawer/content-list-drawer.html
shaundillon: Nope, will remove this

public/components/content-list-drawer/content-list-drawer.html
shaundillon: Good shout. This is more iterative anyway. For self hosted videos, this just wouldn't show

public/components/content-list-drawer/content-list-drawer.js
shaundillon: Possibly, but it's currently not there atoms are being added at a hell of a rate. perhaps something we should look at adding? Ideally, we;d have an 'atom' flag first then only use these to check for the parser to use.

public/lib/atom-service.js
shaundillon: 👍

PR #74, opened 2017-06-21
new audience & fronts team status page

jharewinton:
This adds a new page at /editorialSupport for audience & fronts team staff to display their current availability and status. An API endpoint is also provided for composer to pull in this information in future

combined effort with @rcrphillips


public/editorialsupportteams.js
```diff
@@ -0,0 +1,41 @@
+import 'whatwg-fetch';
+
+function addNewStaff(team) {
+    var name = document.getElementById("name-entry-"+team).value;
+    var endpoint = "/api/editorialSupportTeams?name="+name+"&team="+team;
```
shaundillon: As this is passed through webpack, this can be simplified further to `` `/api.editorialSupportTeams?name=${name}&team=${team}\` `` Check out [template literals](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals)

public/layouts/infopages/info.scss
```diff
@@ -0,0 +1,22 @@
+.admin {
+	padding:30px;
+}
+
+ul.status {
```
shaundillon: the `ul` is unnecessary, this can just be `.status`

public/layouts/infopages/info.scss
```diff
@@ -0,0 +1,22 @@
...
+  list-style-type: none;
+  padding: 0;	
+}
+
+.status label {
```
shaundillon: the `label` here should have its own class and not rely on it being a child of `.status`

public/layouts/infopages/info.scss
```diff
@@ -0,0 +1,22 @@
...
+	margin-left: 5px;
+    font-size: 20px;
+    font-weight: normal;
+}
+
+button.btn-info {
```
shaundillon: `button` not needed here

public/layouts/infopages/info.scss
```diff
@@ -0,0 +1,22 @@
...
+	margin-left: 5px;
+}
+
+footer {
```
shaundillon: add a class for the footer, we should avoid element selectors.

public/editorialsupportteams.js
```diff
@@ -0,0 +1,41 @@
...
+    var endpoint = "/api/editorialSupportTeams/toggle?id="+id+"&active="+active;
+    fetch(endpoint, {
+        method: 'POST',
+        credentials: 'same-origin'
+    }).then(reloadOnComplete)
+}
+
+function updateStatus(id, startText) {
+    var text = document.getElementById("editorialSupportDescription-"+id).value;
+    if (text != startText) {
```
shaundillon: you should use `!==` here, it's more strict as it check type as well as value.

public/editorialsupportteams.js
```diff
@@ -0,0 +1,41 @@
...
+    var name = document.getElementById("name-entry-"+team).value;
+    var endpoint = "/api/editorialSupportTeams?name="+name+"&team="+team;
+    fetch(endpoint, {
+        method: 'PUT',
+        credentials: 'same-origin'
+    }).then(reloadOnComplete)
+}
+
+function reloadOnComplete(response) {
+    if (response.status == 200) {
```
shaundillon: use `===`

rcrphillips:
@shaundillon I think that between myself and @jharewinton we've fixed the review points?

jharewinton:
@paulmr @marialivia16 any chance one of you could take a look at the backend changes in here?

marialivia16:
Looks good to me. Only one question: any reason why we are not using Scanamo to query dynamo?

jharewinton:
@marialivia16 good point. I've not used scanamo before to be honest, so implemented this with the standard sdk as per previous projects. Would be a good thing to do, but I think I will do in a separate PR, as this one needs to go out before the end of next week. That said, its not working on CODE, and VPN + my hotel wifi isn't making it easy to get logs to see why!

PR #75, opened 2017-06-21
Import existing atom from media atom maker

paulmr:
This change allows importing of atoms from the atom editors (currently only media atom maker is supported).

It amends the existing import code so that it checks the URL to determine what tool to import from. To that end, the text in the menu and in the import dialog has been changed to be more generic. The 'Import from' button changes to indicate which tool it has selected for your import.

wf-import-atom


conf/routes
```diff
@@ -53,7 +53,7 @@ GET            /capi/*path                                 controllers.CAPIServi
 
 # Content
 GET            /api/v1/content/:id                         controllers.ContentApi.contentById(id: String)
-
+GET            /api/v1/content/editor/:id                  controllers.ContentApi.contentByEditorId(id: String)
```
marialivia16: It might be good to keep the same naming convention from `workflow`: `/api/atom/:editorId`.

I'm not sure where this v1 api is used.

PR #84, opened 2017-06-28
Fix bug that prevented changing an atom's status from the dropdown

7091lapS:


marialivia16:
[Workflow: Fix status updates](https://trello.com/c/Sp3uz62Q/41-workflow-fix-status-updates)

PR #90, opened 2017-08-15
allow media atom maker in DEV to hit CODE

akash1810:
allow media atom maker DEV to access `/api/sections` and `/api/atom/:editorId`

### DIFF
--- a/app/config/Config.scala
+++ b/app/config/Config.scala
@@ -12,11 +12,15 @@ object Config extends AwsInstanceTags {
   }
   Logger.info(s"running in stage: $stage")
 
-  lazy val domain: String = stage match {
+  def appDomain(appStage: String): String = {
+    appStage match {
       case "PROD" => "gutools.co.uk"
       case "DEV" => "local.dev-gutools.co.uk"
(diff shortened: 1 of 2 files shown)


marialivia16:
Looks good, but this won't hit CODE because of [this](https://github.com/guardian/workflow-frontend/blob/master/app/config/Config.scala#L17) config.

akash1810:
Updated now, its not the best code ever written, but should be ok for now... I've added a `TODO` 😈

app/config/Config.scala
```diff
@@ -28,6 +32,12 @@ object Config extends AwsInstanceTags {
 
   lazy val mediaAtomMakerUrl: String = s"https://video.$domain"
 
+  // TODO do this better!
+  lazy val mediaAtomMakerUrlForCode: String = stage match {
```
marialivia16: Maybe not the best name as it will handle `PROD` as well, but it'll do for now 👀

akash1810: yeah, I couldn't think of anything better

PR #98, opened 2017-09-12
add editor id to filters to query stubs by editor id

Reettaphant:
Allows for passing editorId parameter for the stubs endpoint in workflow. 

Linked to https://github.com/guardian/workflow/pull/1028
https://trello.com/c/rOymNax4/122-open-workflow-to-individual-atom


akash1810:
There's a small bug with this: the left hand filters should be disabled when editorId is on the query string, similar to when looking at an individual composer piece.

akash1810:
PR - https://github.com/guardian/workflow-frontend/pull/99

PR #103, opened 2017-10-18
Support atomworkshop atoms

Reettaphant:
Adds support atom workshop atoms in workflow: all atom workshop atoms in the atom types list can be tracked in workflow and an ability to create new atoms in atom workshop in workflow.  

Adds support for reader question atoms (storyquestions) specifically and a drawer for them.

Depends on:

https://github.com/guardian/workflow/pull/1035
https://github.com/guardian/atom-workshop/pull/170


public/components/stub-modal/stub-modal.js
```diff
@@ -37,7 +37,13 @@ function StubModalInstanceCtrl($rootScope, $scope, $modalInstance, $window, conf
...
     $scope.statuses = statusLabels;
     $scope.cdesks = _wfConfig.commissioningDesks;
-    $scope.atomTypes = ['Media'];
+    $scope.atomTypes = _wfConfig.atomTypes.map(type => {
+      if (type === 'storyquestions') {
+        return {value: type, displayName: 'reader questions'};
+      } else {
+        return {value: type, displayName: type};
+      }
+    });
```
philmcmahon: Could you move this logic into a function to make it a bit more obvious why it's happening ? Maybe `makeAtomTypeDropdownData` or something. Also maybe reader questions should be capitalised?

Reettaphant: Yes, that sounds sensible. I'll capitalise media atoms as well.

PR #105, opened 2017-10-23
Add to indesign button

rcrphillips:
Adds the ability to push a Composer article to InDesign instead of InCopy. The 'seen in incopy' stuff is still useful and relevant even if it's been pushed to indesign, which is why I have kept that contract there. 
screen shot 2017-10-23 at 17 40 36


public/components/content-list-drawer/content-list-drawer.js
```diff
@@ -28,6 +28,13 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
             });
     }
 
+    function buildIndesignUrl(fields) {
+        return config.indesignExportUrl
```
philmcmahon: You could avoid duplicating the logic in the indesign/incopy functions with a something like 
```
   function buildExportUrl(baseExportUrl) {
        return baseExportUrl
            .replace(/\$\{(.*?)\}/g, function(match, fieldName) {
                return fields[fieldName] || "";
            });
}
```

public/app.js
```diff
@@ -76,6 +76,11 @@ angular.module('workflow',
                    "|^\\s*" + _wfConfig.incopyExportUrl.match("^.*?:")[0])
         );
 
+        $compileProvider.aHrefSanitizationWhitelist(
```
philmcmahon: same as below - you could avoid duplicated logic here by having a function which takes the export url as a parameter

philmcmahon:
Looks good! Just a couple of things to fix.

public/app.js
rcrphillips: @philmcmahon Ok, I gave it a shot myself, but need your help with these.

philmcmahon: Sorry, I will help out tomorrow!

rcrphillips:
@philmcmahon could you double check this again?

philmcmahon:
👍

PR #106, opened 2017-10-25
Make parseComposerData return a promise

philmcmahon:
...which will be failed if the composer response doesn't contain at least the content id.

The hope is that this will solve some of the problems that we're getting where content appears in workflow without a composer or editor id associated with it.

This change also removes code relating to 'archived' content (no longer a thing) and an unused $scope.errorMsg variable.

### DIFF
--- a/public/lib/content-service.js
+++ b/public/lib/content-service.js
@@ -92,18 +92,20 @@ angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService
                  */
                 createInComposer(stub, statusOption) {
 
-                    return wfComposerService.create(stub.contentType, stub.commissioningDesks, stub.commissionedLength).then( (response) => {
+                    return wfComposerService.create(stub.contentType, stub.commissioningDesks, stub.commissionedLength)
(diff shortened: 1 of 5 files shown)


public/lib/content-service.js
```diff
@@ -92,18 +92,20 @@ angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService
                  */
                 createInComposer(stub, statusOption) {
 
-                    return wfComposerService.create(stub.contentType, stub.commissioningDesks, stub.commissionedLength).then( (response) => {
+                    return wfComposerService.create(stub.contentType, stub.commissioningDesks, stub.commissionedLength)
+                    .then( (response) => wfComposerService.parseComposerData(response, stub))
+                    .then((updatedStub) => {
 
-                        wfComposerService.parseComposerData(response.data, stub);
+                        // wfComposerService.parseComposerData(response.data, stub);
```
mbarton: remove?

philmcmahon: eep. fail. thanks

PR #107, opened 2017-10-30
Commissioned length

paulmr:
Adds a UI compontent for adding a commissioned length when creating a new article in workflow, and also for editing the commissioned length from the drawer of an existing item.


rcrphillips:
Tested the changes end-to-end, so just needs a code 👍

philmcmahon:
👍

PR #110, opened 2017-11-15
Use Circe

marialivia16:
Look at all the code that can go away!

The interesting bits of this PR are the encoders and decoders for in `Stub.scala` because we are dealing with a flat json structure. 

Other things to keep in mind:
- `circe` doesn't deal with default values by default, we need to import the `circe extra` library to make it work and define an implicit configuration in the companion objects.
- We're using the `enumeratum` library to deal with enums.


marialivia16:
[Migrate to Circe in workflow-frontend](https://trello.com/c/EBsI3w0D/137-migrate-to-circe-in-workflow-frontend)

app/controllers/Api.scala
```diff
@@ -58,8 +56,8 @@ object Api extends Controller with PanDomainAuthActions {
     }
 
     CommonAPI.getStubs(qs).asFuture.map {
-      case Left(err) => InternalServerError
-      case Right(contentResponse) => Ok(Json.toJson(contentResponse))
+      case Left(_) => InternalServerError
+      case Right(contentResponse) => Ok(contentResponse.asJson.spaces4)
```
philmcmahon: Should this maybe just be without pretty spaces? Since the app/browser can handle that and it might slightly reduce the response size? I don't know if it matters hugely though.

app/controllers/Api.scala
```diff
@@ -290,30 +252,28 @@ object Api extends Controller with PanDomainAuthActions {
   def statusus = CORSable(defaultCorsAble)  {
     APIAuthAction.async { implicit req =>
       for(statuses <- StatusDatabase.statuses) yield {
-        Ok(renderJsonResponse(statuses))
+        Ok(renderJsonResponse(statuses).asJson.spaces4)
```
philmcmahon: likewise here (see previous comment)

common-lib/src/main/scala/models/DateFormat.scala
```diff
@@ -0,0 +1,22 @@
...
+  implicit val dateTimeDecoder = new Decoder[DateTime] {
+    def apply(c: HCursor): Decoder.Result[DateTime] = {
+      val dateTime = for {
+        value <- c.focus
+        string <- value.asString
+      } yield new DateTime(string)
+      dateTime.fold[Decoder.Result[DateTime]](Left(DecodingFailure("could not decode date", c.history)))(dt => Right(dt))
+    }
+  }
+}
```
philmcmahon: Nice :) this code is probably a good reason to keep this repo open source so it can be copy pasted by others!

common-lib/src/main/scala/models/Stub.scala
```diff
@@ -25,167 +60,50 @@ case class Stub(id: Option[Long] = None,
...
-
-  implicit val stubWrites: Writes[Stub] = Json.writes[Stub]
-
+  implicit val customConfig: Configuration = Configuration.default.withDefaults
+
+  implicit val encoder: Encoder[Stub] = deriveEncoder
+
+  implicit val decoderHelper: Decoder[Stub] = new Decoder[Stub] {
+    def apply(c: HCursor): Result[Stub] = {
+      val sectionName: String = c.downField("section").downField("name").as[String].getOrElse(c.downField("section").as[String].getOrElse(""))
```
philmcmahon: maybe it's worth adding a comment to explain why this is necessary

common-lib/src/main/scala/models/Utils.scala
```diff
@@ -1,27 +0,0 @@
...
-  }
-
-  implicit def enumWrites[E <: Enumeration]: Writes[E#Value] = new Writes[E#Value] {
-    def writes(v: E#Value): JsValue = JsString(v.toString)
-  }
-
-  implicit def enumFormat[E <: Enumeration](enum: E): Format[E#Value] = {
-    Format(enumReads(enum), enumWrites)
-  }
-}
```
philmcmahon: 👍 💀

common-lib/src/main/scala/models/WorkflowContent.scala
```diff
@@ -1,199 +1,24 @@
...
-        c => c.getOrElse(0)
-      } ~
-      (__ \ "launchScheduleDetails").readNullable[LaunchScheduleDetails].map(_.getOrElse(LaunchScheduleDetails(None, None, false))) ~
-      (__ \ "statusFlags").read[WorkflowContentStatusFlags]
-     )(WorkflowContent.apply _)
-}
-
-case class ContentItemIds(stubId: Long, composerId: Option[String])
-object ContentItemIds {
-  implicit val jsonFormat: Format[ContentItemIds] = Json.format[ContentItemIds]
```
philmcmahon: delete gif

app/controllers/Api.scala
marialivia16: Good point. I got too excited about that 😆 I'll change it!

PR #129, opened 2018-03-01
add GA

shaundillon:
This currently covers the original mixpanel tracking (Removed in https://github.com/guardian/workflow-frontend/commit/990e12b63191fa8bf4c6aa9da34cc957bcfda881), we can always add more now the GA service is in.

### DIFF
--- a/public/lib/analytics.js
+++ b/public/lib/analytics.js
@@ -1,0 +1,49 @@
+/**
+ * Analytics module which listens for events and tracks them as appropriate
+ * within Google analytics.
+ */
+
+import angular from 'angular';
+
+const initGA = gaId => {
+    // tracking script should be on the page already
+    if (gaId) {
+        window.ga =
+            window.ga ||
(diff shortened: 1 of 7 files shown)


public/lib/analytics.js
```diff
@@ -0,0 +1,50 @@
...
+
+
+const wfAnalyticsServiceMod = angular.module('wfAnalyticsServiceMod', []);
+
+wfAnalyticsServiceMod.service('wfAnalyticsService', [
+    '$rootScope',
+    function($rootScope) {
+
+        //setup ga
+        const gaId = 'UA-78290349-2';
```
mbarton: Is it safe to put this in an open repo?

shaundillon: It was only supposed to be here temporarily but I forgot to move it. It's not unsafe as you can find it out from the network tab on any page anyway, and all anyone can do is send analytics to our account, but it's not ideal in terms of keeping config in one place. I'll move it over.

PR #130, opened 2018-03-01
Group fronts in editorial support page

mbarton:
The editorial support page for fronts team has got rather out of hand. It was designed using the same data model of a "Team", with the description for each person indicating the front they are currently covering. Given the number of people and fronts it would be simpler for them to simply update the name against the front.

I have moved the backing store to a [cloud-formed table](https://github.com/guardian/editorial-tools-platform/pull/146) and switched from using randomly generated IDs to an ID indicating the team and name. The frontend then has special code to render fronts even though the backing data is in the same format as the other teams.

If a user is editing multiple fronts at the same time they simply enter their name against both. We group the results by user for display in Composer eg "US Front, Intl Front".

If a user deletes their name the front remains in the list but will not display in Composer.

I've switched to plain HTML forms rather than JS on the front end, using hidden inputs for `id` and `team` as appropriate. This also allows us to have a single end-point handling adds and updates.

I've removed the ability to delete users. It's an open question about whether it's worth putting back?

49494a7fbcc1bcfcbb418f47d2aceb01

Net deletion of code 💥 

TODO:
- [x] test in CODE
- [x] deploy cloudformation for new table in PROD
- [x] migrate existing data to new table
- [ ] deploy to PROD


philmcmahon:
Lovely. 👍

PR #140, opened 2018-05-24
Mlc fix commissioned length

marialivia16:
This solves 2 issues:
- When commissioned length was deleted it was sent as an empty string `""` to composer which caused the content to become unstreamable meaning CAPI was loosing updates. We need to send a DELETE request instead.
- When the value is updated we should also save it in the workflow db.

Depends on: https://github.com/guardian/workflow/pull/1048

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -176,6 +176,16 @@ object Api extends Controller with PanDomainAuthActions {
     )}
   }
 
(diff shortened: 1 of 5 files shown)


tomrf1:
👍 looks good!

marialivia16:
Thanks for reviewing, @tomrf1 ! Can you please have a quick look at this as well https://github.com/guardian/workflow/pull/1048 ? We need to get that one in first.

PR #142, opened 2018-06-25
Apply styles to broken date pickers

jonathonherbert:
Re-adds a missing stylesheet to the date-time-picker components, which currently look a bit odd, and makes a few tweaks to present them more consistently.


public/components/content-list-drawer/_content-list-drawer.scss
```diff
@@ -2,7 +2,9 @@
 .drawer {
     display: flex;
     width: 100%;
-
+    * {
+        box-sizing: border-box;
```
shaundillon: 🙌

PR #144, opened 2018-08-07
rename Atoms in the "create new" dropdown

franziskas:
We've received the following request:

> Using the "Create new Atom" function in Workflow creates a video atom. This has led to confusion as people expect it to go to the Atom menu from where you can choose timeline/quick guide etc.
> 
> Could this be renamed "Create Video Atom" or better link to the Atom dashboard?

Ultimately, we probably want to link to https://atomworkshop.code.dev-gutools.co.uk/create and get rid of this form or allow more atoms to be created without leaving workflow. This is just a quick fix until we have time to properly investigate.
Before:
screen shot 2018-08-07 at 12 27 21

After:
screen shot 2018-08-07 at 12 25 37

### DIFF
--- a/public/lib/content-service.js
+++ b/public/lib/content-service.js
@@ -22,7 +22,7 @@ angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService
                         "gallery": "Gallery",
                         "interactive": "Interactive",
                         "picture": "Picture",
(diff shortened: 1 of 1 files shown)


akash1810:
> This is just a quick fix until we have time to properly investigate.

👍

PR #147, opened 2018-09-06
Send production office to composer

marialivia16:
Production office wasn't sent to Composer and it was always defaulting to `UK`


marialivia16:
[Production office for an article created in workflow is not synchronised to Composer](https://trello.com/c/jxyQvHF9/379-production-office-for-an-article-created-in-workflow-is-not-synchronised-to-composer)

philmcmahon:
Oh nice!! Can't believe this has been broken so long 🇬🇧 🇺🇸 🇦🇺 :)

PR #150, opened 2018-09-25
rm about page

akash1810:
@paulmr and @philmcmahon aren't in Tools anymore, so [this page](https://workflow.gutools.co.uk/about) is outdated...


franziskas:
https://github.com/guardian/workflow-frontend/blob/aa-rm-about/public/layouts/dashboard/dashboard-user.html

PR #154, opened 2018-10-02
Delete subscriptions

mbarton:
A page to delete existing subscriptions. This is the minimum UI needed to send it out for a PoC amongst some users:

e5a6e1c889a17f7cd57946d12d8e1ac4

There's also feedback on the button when you click subscribe:

51ba7861894a721f0a4c809072a7f3cf

I've also fixed a problem where users could only subscribe to one query at once. This was because we were only hashing the endpoint to construct the subscription ID. Now we hash both the endpoint and query. This means users cannot create duplicate subscriptions.

### DIFF
--- a/common-lib/src/main/scala/models/Subscription.scala
+++ b/common-lib/src/main/scala/models/Subscription.scala
@@ -10,14 +10,18 @@ import io.circe.parser.decode
 import io.circe.syntax._
 import io.circe.{Decoder, Encoder}
 
+import play.api.data.Forms._
+import play.api.data._
+
 case class SubscriptionKeys(p256dh: String, auth: String)
 case class SubscriptionEndpoint(endpoint: String, keys: SubscriptionKeys)
 // seenIds is optional as it encodes three states:
 //   None          -> we have not seen any content under the given query yet so don't fire notifications
 //   Some(Nil)     -> we last saw no content matching the query
 //   Some(content) -> the content we saw last (ie do a diff and fire notifications)
-case class Subscription(query: Subscription.Query, seenIds: Option[Set[Long]], endpoint: SubscriptionEndpoint)
+case class Subscription(email: String, query: Subscription.Query, seenIds: Option[Set[Long]], endpoint: SubscriptionEndpoint)
 case class SubscriptionUpdate(title: String, body: String, url: Option[String])
+case class DeleteSubscription(id: String)
 
 object Subscription {
   implicit val customConfig: Configuration = Configuration.default.withDefaults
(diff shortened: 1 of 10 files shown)


common-lib/src/main/scala/models/Subscription.scala
```diff
@@ -36,18 +40,37 @@ object Subscription {
 
   type Query = Map[String, Seq[String]]
 
-  def id(endpoint: SubscriptionEndpoint): String = {
+  val form = Form(
+    mapping("id" -> text)
+    (DeleteSubscription.apply)(DeleteSubscription.unapply)
```
akash1810: 🤔 what's this doing?

public/layouts/dashboard/dashboard-create.js
```diff
@@ -20,9 +20,16 @@ angular
         };
 
         $scope.subscriptionsSupported = subscriptionsSupported();
+        $scope.subscriptionStatus = null;
+
         $scope.registerSubscription = () => {
-            // TODO MRB: show spinner and success or similar
-            registerSubscription();
+            $scope.subscriptionStatus = "Subscribing...";
```
akash1810: minor: this could be a `bool` in the controller, allowing the view to remain in charge of what's rendered

public/layouts/dashboard/dashboard-user.html
```diff
@@ -34,8 +34,21 @@
         </li>
 
         <li class="dropdown-toolbar__item" ng-if="subscriptionsSupported">
-            <a class="dropdown-toolbar__item-label" href="#" ng-click="registerSubscription()">
-                <span class="dropdown-toolbar__item-title" title="">Subscribe to this page</span>
+            <span class="dropdown-toolbar__item-label" href="#" ng-click="registerSubscription()" ng-if="subscriptionStatus">
```
akash1810: the semantic web is 😢

conf/routes
```diff
@@ -13,6 +13,8 @@ GET            /dashboard                                  controllers.Applicati
 GET            /about                                      controllers.Assets.at(path="/public", file="/static/about.html")
 GET            /editorialSupport                           controllers.Application.editorialSupport
 POST           /editorialSupport                           controllers.Application.updateEditorialSupport
+GET            /subscriptions                              controllers.Notifications.subscriptions
+POST           /subscriptions                              controllers.Notifications.deleteSubscription
```
akash1810: are we missing a `DELETE` route?

mbarton: Irritatingly you can't do `method="DELETE"` in an HTML form. I've worked around this in the editorial support page stuff by adding a hidden input `action=UPDATE` or `action=DELETE` but right now we can only delete so didn't see any reason to add it.

common-lib/src/main/scala/models/Subscription.scala
mbarton: Just copy paste of how we're using the Play Forms DSL in the editorial support page. It hydrates a case class from the given form data.

Since we just have an ID I'll take a look and see if we don't need the case class.

public/layouts/dashboard/dashboard-user.html
mbarton: Oops it shouldn't have the `ng-click` handler if it's currently subscribing

public/layouts/dashboard/dashboard-create.js
mbarton: It could and probably should be but I'm going to leave it to avoid too much copy-pasting in the template (two `ng-if` statements on seperate `spans`)

common-lib/src/main/scala/models/Subscription.scala
mbarton: Reading the play docs it seems they encourage the use of case classes (with decent reasons for any forms that aren't as basic as mine). I'll leave it I think.

https://www.playframework.com/documentation/2.6.x/ScalaForms

PR #155, opened 2018-10-06
Only show user subscriptions after deletion

mbarton:
Fix silly bug where the user would be shown all the subscriptions in the Manage Subscriptions page after they deleted one, not just their own.


LATaylor-guardian:
👍

PR #158, opened 2018-10-16
Add chart atom type to workflow.

susiecoleman:
screen shot 2018-10-16 at 09 33 19

screen shot 2018-10-16 at 09 33 46


public/components/content-list-drawer/content-list-drawer.html
```diff
@@ -334,6 +334,52 @@
                         </li>
                     </ul>
                 </div>
+                
+                <!--Charts ATOM DRAWER-->
```
Reettaphant: Since most of this template is the same as the story-questions maybe we could duplicate less code by wrapping the individual parts that are different into ng-ifs? If we could move the `getAtomDisplayName` somewhere where we can share it with this editor then the Content type could come from this function and we'd only have the last list in this template that is different between the two atom types.

susiecoleman: Updated the code to make it more general

PR #161, opened 2018-10-23
Replace old form links with health form

marialivia16:


marialivia16:
[Change Composer feedback links to point to health feedback form](https://trello.com/c/e1bXMXga/531-change-composer-feedback-links-to-point-to-health-feedback-form)

PR #164, opened 2018-12-27
WIP: Phrase subscriptions

mbarton:
Extends Workflow notifications to fire when certain phrases are used in content.

TODO
- [x] CRUD UI for managing subscriptions
- [ ] Turn the `notifications` project into an app and consume CAPI events
   - Change existing behaviour to trigger from SNS updates?
   - Retain seen in content time per sub to avoid spamming notifications on every save for the same phrase
    - Cache subscriptions (update every N invocations?) to avoid heavy dynamo I/O
- [ ] Add [event source mapping](https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html) trigger in Cloudformation

### DIFF
--- a/app/Global.scala
+++ b/app/Global.scala
@@ -1,12 +1,15 @@
+import java.security.Security
+
 import config.Config
 import lib.{LogConfig, LoggingFilter, RedirectToHTTPSFilter}
+import org.bouncycastle.jce.provider.BouncyCastleProvider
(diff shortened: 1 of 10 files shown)


mbarton:
Closing for now. After discussing with @jonathonherbert IRL we agreed that it makes sense for subscriptions to live in the main workflow codebase as Prole is already consuming the CAPI firehose there

PR #166, opened 2019-02-26
Expose priority as an API for Composer

mbarton:
https://github.com/guardian/flexible-content/pull/3359

Also added bonus logging of errors when getting content since it was obstructing me debugging why my local build was running (I had the wrong port in my config)

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -4,7 +4,7 @@ import cats.syntax.either._
 import com.gu.pandomainauth.action.UserRequest
 import com.gu.workflow.api.{ApiUtils, CommonAPI, PrototypeAPI, SectionsAPI}
 import com.gu.workflow.lib.DBToAPIResponse.getResponse
-import com.gu.workflow.lib.StatusDatabase
+import com.gu.workflow.lib.{Priorities, StatusDatabase}
 import com.gu.workflow.util.SharedSecretAuth
 import config.Config
 import config.Config.defaultExecutionContext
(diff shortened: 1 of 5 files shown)


app/controllers/Api.scala
```diff
@@ -57,8 +57,12 @@ object Api extends Controller with PanDomainAuthActions with SharedSecretAuth {
     val qs: Map[String, Seq[String]] = queryString(req)
 
     CommonAPI.getStubs(qs).asFuture.map {
-      case Left(_) => InternalServerError
-      case Right(contentResponse) => Ok(contentResponse.asJson.noSpaces)
+      case Left(err) =>
```
sihil: Without opening an IDE I can't tell if this is an exception. If so then it should be in the second param in the logger line. I'm guessing it's a string though.

app/controllers/Api.scala
```diff
@@ -282,6 +286,12 @@ object Api extends Controller with PanDomainAuthActions with SharedSecretAuth {
     }
   }
 
+  def priorities = CORSable(atomCorsAble) {
+    AuthAction {
```
sihil: Should this be an `APIAuthAction`?

app/controllers/Api.scala
mbarton: It's an [ApiError](https://github.com/guardian/workflow-frontend/blob/master/app/lib/Response.scala#L8) wrapper object around the HTTP response from the Workflow datastore app

app/controllers/Api.scala
mbarton: Agreed, all the metadata endpoints should be (`statuses`, `sections`, `allowedAtomTypes` and `priorities`).

I've raised https://github.com/guardian/workflow-frontend/pull/169 for it

PR #170, opened 2019-03-25
allow atom tools to hit priority endpoint

akash1810:
CORS is fun

We need to explicitly allow access from Atom Tools, similar to the [status update endpoint](https://github.com/guardian/workflow-frontend/blob/master/app/controllers/Api.scala#L173)


ajwl:
oh great thanks Akash

PR #176, opened 2019-10-18
Remove storyquestions

regiskuckaertz:


ajwl:
I presume this is because "story questions" is being withdrawn as an atom type?

regiskuckaertz:
Yes

PR #180, opened 2020-03-12
Add opening in incopy if the article is already linked

itsibitzi:
@blishen @hoyla @paperboyo (can someone add Mariana to the org please?)

If the article is already linked in incopy then we want to be able to open it instead of sending it again which is a far less common operation.

Do not merge! Relies on a fast script which doesn't yet exist!!

Old export URL is like so: `gnm://composer/export/${composerId}`

Currently the new open URL will be like so `gnm://composer/open/${composerId}`, I can change the composer ID to be the bundle ID if that's preferred?

Before - weird and rarely used.
image

After - super handy and makes you cool.
image


itsibitzi:
Updated with the new URL for opening in InCopy, also removed the open in InDesign feature as this is a little more complex on the fast scripts side.

app/config/Config.scala
```diff
@@ -63,6 +63,7 @@ object Config extends AwsInstanceTags {
 
   lazy val sharedSecret: String = config.getConfigStringOrFail("api.sharedsecret")
 
+  lazy val incopyOpenUrl: String = "gnm://openinincopy/${storyBundleId}/checkout/readwrite"
```
paperboyo, line 66: This is incorrect, it should read:
`gnm://openinincopy/storybundle/${storyBundleId}/checkout/readwrite`
cc @hoyla

itsibitzi:
This can now be merged.

public/components/content-list-drawer/content-list-drawer.js
```diff
@@ -174,8 +167,18 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
                 }
 
                 // TODO: move build incopy URL to decorator
-                $scope.incopyExportUrl = buildExportUrl({ "composerId": contentItem.composerId }, config.incopyExportUrl);
-                $scope.indesignExportUrl = buildExportUrl({ "composerId": contentItem.composerId }, config.indesignExportUrl );
+
+                $scope.buildUrl = function(fields, url) {
+                    console.log(url, fields)
```
jonathonherbert, line 172: Still needed?

itsibitzi: Ah no! Apologies :)

PR #185, opened 2020-04-29
adding check that $JAVA_HOME is set

srbd:
- otherwise can result in null pointer exceptions

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -4,7 +4,7 @@ red='\x1B[0;31m'
 
 if [[ -z "$JAVA_HOME" ]]; then
     echo -e "\r${red}JAVA_HOME must be set" 1>&2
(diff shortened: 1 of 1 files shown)


scripts/setup.sh
```diff
@@ -2,6 +2,11 @@
 
 red='\x1B[0;31m'
 
+if [[ -z "$JAVA_HOME" ]]; then
+    echo -e "\r${red}JAVA_HOME must be set" 1>&2
```
akash1810, line 6:
```suggestion
    echo -e "\r${red}JAVA_HOME must be set" 1>&2
    echo "This can be done by adding \"export JAVA_HOME=$(/usr/libexec/java_home)\" to your ~/.profile"
```

srbd: yep, sounds good 👍

PR #189, opened 2020-05-01
Use a java version which does not have a known bug!

aug24:
Per S Hildrew email:

242 causes errors that look like
```
Caused by: java.lang.NullPointerException
    at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1847)
    at java.lang.Runtime.loadLibrary0(Runtime.java:871)
    at java.lang.System.loadLibrary(System.java:1124)
```

Turns out that OpenJDK fixed a 'bug' that lots of people (Play included) relied on in 1.8.0_242.  You can see the resulting chaos at https://bugs.openjdk.java.net/browse/JDK-8240521.


.java-version
```diff
@@ -1 +1 @@
-1.8.0.242
+1.8.0.202
```
aug24, line 1:
```suggestion
1.8.0.232
```

aug24:
> Can we add more detail here? What is the bug exactly? Is 202 the latest stable version that Workflow runs in? Most other apps are ok on 232 IIRC.

Added detail, swapped to 232.

PR #190, opened 2020-05-01
Add print word count

aug24:
Composer now supports marking text as webonly or printonly, and sending the resulting different word counts to workflow (see https://github.com/guardian/flexible-content/pull/3478)

This PR provides workflow with the facility to read from workflow backend and display those values in an optional column.

### DIFF
--- a/public/components/content-list/_content-list.scss
+++ b/public/components/content-list/_content-list.scss
@@ -48,6 +48,7 @@ $textualIconSize: 22px;
             &--links,
             &--published-state,
             &--wordcount,
+            &--printwordcount,
             &--commissionedLength,
             &--needsLegal {
                 padding: 15px 10px 5px 5px;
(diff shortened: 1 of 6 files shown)


AWare:
https://github.com/guardian/flexible-content/pull/3478
https://github.com/guardian/workflow-frontend/pull/190
https://github.com/guardian/workflow/pull/1055

These three PRs all depend on each other.

public/components/content-list/_content-list.scss
```diff
@@ -114,7 +115,7 @@ $textualIconSize: 22px;
                 width: 40px;
             }
 
-            &--wordcount {
+            &--printwordcount {
```
sihil, line 118: Slightly puzzled as to why the `wordcount` was duplicated originally?

aug24: Mistake, I expect.  I took the opportunity to cleanly rectify it.

PR #191, opened 2020-05-07
remove RedirectToHTTPSFilter as no longer needed

akash1810:
We've set the HSTS header on the entire domain, so there is no way of connecting via http now, so we no longer need this filter.


akash1810:
Tested on CODE.

PR #192, opened 2020-05-07
use Permissions for permissions

akash1810:
Rather than roll our own permissions configuration, use the Permissions service. This allows us to have a single place to manage permissions, keeping things simple.

Requires https://github.com/guardian/permissions/pull/123.


project/Dependencies.scala
```diff
@@ -57,4 +57,11 @@ object Dependencies {
   val webPushDependencies = Seq(
     "nl.martijndwars" % "web-push" % "3.1.1"
   )
+
+  // bump this once new permissions client is available
+  val permissionsClientVersion = "2.0"
```
akash1810, line 62: Requires https://github.com/guardian/permissions/pull/123.

PR #193, opened 2020-05-11
Setup tweaks

jennygrahamjones:
Some small changes to the setup script and installation commands in the README, based on my experience with setting up workflow-frontend locally for the first time.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -17,13 +17,12 @@ if [ $? != "0" ]; then
     exit 1
 fi
 
-source "$NVM_DIR/nvm.sh"
-
 if ! test -e "$NVM_DIR/nvm.sh"; then
     echo -e "NVM not found. NVM is required to run this project"
     echo -e "Install it from https://github.com/creationix/nvm#installation"
     exit 1
   else
+    source "$NVM_DIR/nvm.sh"
     nvm install
   fi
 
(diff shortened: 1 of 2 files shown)


scripts/setup.sh
```diff
@@ -17,6 +17,16 @@ if [ $? != "0" ]; then
     exit 1
 fi
 
+source "$NVM_DIR/nvm.sh"
+
+if ! test -e "$NVM_DIR/nvm.sh"; then
```
SHession, line 22: Am I right in thinking this checks for the existence of nvm.sh after already executing it?

jennygrahamjones: Oops. That line is in the wrong place 😅

jennygrahamjones:
@mbarton is there anything additional you think is worth adding?

mbarton:
I wonder if it's worth documenting what bits are "expected" not to work for now, both to help anyone in the future and so we can check them off as and when we fix them. I think there's:

- CAPI
- Presence
- Assign to me

Preferences is probably in there as well. I've managed to run workflow-frontend locally without it and saw no errors.

Let's also remove the `NB At present, the Play Framework version does not support https as well as it should.` bit, not sure what it's alluding to

akash1810:
I'd like to get this tool in line with others, in that, in DEV, it should call out to CODE services and not require one to run them (preferences, presence etc) locally. Happy to help!

PR #194, opened 2020-05-12
Display web and print word count in Furniture panel

jennygrahamjones:
Continues #190 by adding the print word count to the Furniture panel in Workflow. 

I'd love to name the CSS class `wordcount` something more generic, if anyone has any suggestions.

Before: 
Screenshot 2020-05-12 at 11 41 56

After:
Screenshot 2020-05-12 at 11 41 15

### DIFF
--- a/public/components/content-list-drawer/content-list-drawer.html
+++ b/public/components/content-list-drawer/content-list-drawer.html
@@ -131,7 +131,7 @@
                             </div>
                             <div class="drawer__item-wordcount">
                                 <span class="drawer__item-title">Print word count</span>
-                                <span class="drawer__item-content">{{ contentItem.printWordCount }} </span>
+                                <span class="drawer__item-content">{{ contentItem.printWordCount || "Unknown" }} </span>
                             </div>
                         </li>
                         <li class="drawer__item" ng-show="contentItem.contentType == 'article'">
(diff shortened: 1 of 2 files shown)


public/components/content-list-drawer/content-list-drawer.html
```diff
@@ -131,7 +131,7 @@
                             </div>
                             <div class="drawer__item-wordcount">
                                 <span class="drawer__item-title">Print word count</span>
-                                <span class="drawer__item-content">{{ contentItem.printWordCount }} </span>
+                                <span class="drawer__item-content">{{ contentItem.printWordCount || "Unknown" }} </span>
```
jennygrahamjones, line 134: @SHession this should fix what we were discussing earlier :)

SHession: It might be worth adding that same bit to the line 130, just incase wordCount ever comes back as null

jennygrahamjones: I believe this should be handled already. At least, it is in my local.

Screenshot 2020-05-13 at 15 30 12

I think this value is pulled, along with some other "Unknown"s for headline, standfirst, etc, from the [`emptyCapiContentObject`](https://github.com/guardian/workflow-frontend/blob/c2f6a548877b30f5059bfce037d14ccb9c32ae3b/public/lib/capi-content-service.js#L71).

SHession: Oh yeah, you're right. In which case it might be worth changing _content-list-item.js :91_ instead. That's a bit pedantic though, so if you don't want to change it that's fine.

PR #195, opened 2020-05-12
Last modified column

SHession:
"_It's not easy to see at a glance when an article has most recently been modified in Composer, from Workflow, nor the last person to touch it._" - [Trello](https://trello.com/c/4zpbltH1/88-expose-workflows-last-modified-as-a-column-off-by-default)

This PR adds a **Last Modified** column to the workflow display, this should make it easier to see what content has been edited most recently. The column displays the same value that is currently visible in the management panel.

image

### How to review and test

Open workflow, select the column options button (top right column heading icon), add the **Last Modified** column. It should now show the value where available.

### DIFF
--- a/public/components/content-list-item/_content-list-item.scss
+++ b/public/components/content-list-item/_content-list-item.scss
@@ -232,7 +232,8 @@
 
         &--deadline,
         &--created,
-        &--legal {
+        &--legal,
+        &--last-modified {
             width: 100px;
         }
 
@@ -259,12 +260,14 @@
         }
 
         &--section,
-        &--deadline {
+        &--deadline,
+        &--last-modified {
             @extend %fs-data-1;
         }
 
         &--deadline,
-        &--created {
+        &--created,
+        &--last-modified {
             white-space: nowrap;
         }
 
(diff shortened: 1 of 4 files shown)


paperboyo:
👏 
Just a point of discussion: if we are to keep this formatting of date (as oppose to choosing, I think, more concise and more readable in column context, if less consistent with details view, `12/05/2020 11:46`), we should make sure it doesn’t sort by `Tue` :-). (see Questions in Trello and a mockup that, wrongly, wraps onto two rows)

SHession:
> if we are to keep this formatting of date (as oppose to choosing, I think, more concise and more readable in column context, if less consistent with details view, `12/05/2020 11:46`)

@paperboyo I went with the formatting that's being used for the deadline column. If you would rather change it, we should aim to be consistent with our dates across the columns. Additionally, when it comes to sorting, we should be able to sort by the value itself rather than the displayed text. 

Would you still like me to change the date formats in the columns?

paperboyo:
@SHession As long as it’s possible to to sort correctly, this can go in. Unless the team feels another formatting would be more readable.

SHession:
@akash1810 I'm not 100% on that one. I've used the same property as the one displayed in the management panel so it should be consistent, but I don't know exactly how it's populated. I wasn't able to see changes in my local composer reflected in my local Workflow, I would be interested in setting this up so I could test it further. I believe for this PR it should be adequate.

PR #199, opened 2020-05-14
Display n/a when no print count found

jennygrahamjones:
Continues https://github.com/guardian/workflow-frontend/pull/194

Currently, when there is no print word count, "Unknown" is displayed. In column view, if there is no print word count, nothing is displayed. Instead of displaying the potentially confusing "Unknown", which mirrors what we would display if we cannot retrieve word count data from CAPI, this PR suggests that we display "n/a":

Screenshot 2020-05-14 at 17 27 17

This styling mirrors how we display other fields in the Furniture details view that do not have a relevant value (see Standfirst, Main media, and Caption below):

Screenshot 2020-05-14 at 17 29 14

Alternatively, we could use the word "None" here too, but I think that may be potentially misleading for users too, as it suggest there is no word count for print.

Screenshot 2020-05-14 at 17 26 28

### DIFF
--- a/public/components/content-list-drawer/content-list-drawer.html
+++ b/public/components/content-list-drawer/content-list-drawer.html
@@ -131,7 +131,8 @@
                             </div>
                             <div class="drawer__item-wordcount">
                                 <span class="drawer__item-title">Print word count</span>
-                                <span class="drawer__item-content">{{ contentItem.printWordCount || capiData.wordCount }} </span>
+                                <span class="drawer__item-content" ng-class="{'drawer__item-content--empty': !contentItem.printWordCount}">
+                                    {{ contentItem.printWordCount || "n/a" }} </span>
                             </div>
                         </li>
                         <li class="drawer__item" ng-show="contentItem.contentType == 'article'">
--- a/public/components/content-list-drawer/content-list-drawer.html
+++ b/public/components/content-list-drawer/content-list-drawer.html
@@ -131,7 +131,7 @@
                             </div>
                             <div class="drawer__item-wordcount">
                                 <span class="drawer__item-title">Print word count</span>
-                                <span class="drawer__item-content">{{ contentItem.printWordCount || "Unknown" }} </span>
+                                <span class="drawer__item-content">{{ contentItem.printWordCount || capiData.wordCount }} </span>
                             </div>
                         </li>
                         <li class="drawer__item" ng-show="contentItem.contentType == 'article'">


public/components/content-list-drawer/content-list-drawer.html
```diff
@@ -131,7 +131,7 @@
                             </div>
                             <div class="drawer__item-wordcount">
                                 <span class="drawer__item-title">Print word count</span>
-                                <span class="drawer__item-content">{{ contentItem.printWordCount || "Unknown" }} </span>
+                                <span class="drawer__item-content">{{ contentItem.printWordCount || capiData.wordCount }} </span>
```
akash1810, line 134: Should we perform this getOrElse on the server?

jennygrahamjones: @SHession mentioned on https://github.com/guardian/workflow-frontend/pull/194 that we might do the check elsewhere, e.g. in [content-list-item.js](https://github.com/guardian/workflow-frontend/blob/ebf991a393a8355b60c06f1addb2a0db27ef3eb0/public/components/content-list-item/content-list-item.js#L91). Is this what you were thinking?

akash1810: We're rendering print word count in a few places, I'd imagine we'd want to fallback to the web word count in all cases? If so, we can likely assert that if print count is None, web count is a sensible fallback. Doing this on the server keeps the logic in one place and the client thinner.

akash1810: Discussed IRL: falling back to word count suggest that the content has been approved to go through the print process, which I don't think is quite correct. `Unknown` isn't quite correct here either as it implies we tried and failed to get a value. Maybe `n/a` or `None` would be better. Other fields show a grey `None`, so it might make sense to lean on that UX?

jonathonherbert: That sounds sensible 👍

jennygrahamjones: I've updated the PR description with screenshots of the two options. I've gone with "n/a" as I think contextually it makes more sense.

paperboyo: Alternatively, we could show web count greyed out. Because it’s not that until there are some print/web marks, the article will print Unknown number of words in a paper. Or, that it will be n/a there :-)

paperboyo: Also, does this apply to column view too?

jennygrahamjones: > Alternatively, we could show web count greyed out. 

We could, although having thought about it I worry this would be a bit misleading, for the reasons Akash mention above. This is where an expert UX opinion would be wonderful!

> Also, does this apply to column view too?

No, this change only affects details. Column view continues to work as it already does. In production, this means that some stories have an empty print column, while others have matching values in web and print.

paperboyo: _some stories have an empty print column, while others have matching values in web and print_

Oh! I assumed those empty ones are from before the change! So what’s the difference between those with matching and those without?

jennygrahamjones: It looks like it's the difference between content that has and hasn't been updated since we rolled out the original introduction of print word count.

PR #200, opened 2020-05-14
Use CODE Preferences

akash1810:
This allows us to develop Workflow locally w/out needing to run Preferences locally too and use CODE Preferences instead. This will produce a better developer experience.

Requires https://github.com/guardian/editorial-preferences/pull/33.


common-lib/src/main/scala/com/gu/workflow/util/Stage.scala
```diff
@@ -0,0 +1,30 @@
+package com.gu.workflow.util
+
+sealed trait Stage {
```
akash1810, line 3: `String` type is evil 😈

akash1810:
tested on CODE ✅

PR #203, opened 2020-05-14
track ci steps in vcs

akash1810:
We've had a few builds fail to run `yarn` but not exit 0 and so proceeded to the sbt build and deployed with riffraff. This caused the app to be served without any client side assets, rendering it unusable.

Currently TeamCity is responsible for installing yarn and it doesn't look especially stable. Rather than have TeamCity setup to run multiple different steps, we can simplify it by running one step which is to invoke `./scripts/ci.sh` giving us more control.

Invoking a script in the repo also makes it easier to tweak what CI does across branches.

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -e
+
+setupNvm() {
+  if ! test -e "$NVM_DIR/nvm.sh"; then
+    echo -e "NVM not found. NVM is required to run this project."
+    exit 1
+  else
+    source "$NVM_DIR/nvm.sh"
+    nvm install
+  fi
+}
+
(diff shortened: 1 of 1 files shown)


scripts/ci.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo -e "NVM not found. NVM is required to run this project."
+    exit 1
+  else
+    source "$NVM_DIR/nvm.sh"
+    nvm install
+  fi
+}
+
+buildJs() {
+  echo "##teamcity[compilationStarted compiler='webpack']"
```
akash1810, line 16: this creates a group (?) in the build log

image

jonathonherbert:
works great locally (not that we care 😁 ), and I can see the build is happy! 👍

PR #212, opened 2020-05-18
Splits preview link from open in column links

SHession:
The "Open In" column icons are seldom used except preview, which is generally not known about. The icons are taking up valuable screen space when they aren't needed. This PR splits out the "Preview" icon into a separate column. [Trello card](https://trello.com/c/VWMAO4Rb/1879-split-preview-link-from-openin-column)

image

The advantages of this are better use of available screen space and an additional column heading which labels the icon more clearly. 

I have also increased the size of the preview icon for good measure. 

### How to review and test

Open the Workflow dashboard, enable the "Preview" column, observe the correct link and display behaviour.


public/components/content-list-item/templates/preview.html
```diff
@@ -0,0 +1,8 @@
+<td class="content-list-item__field--preview">
```
akash1810, line 1: I'd keep the `td` co-located with it's parent `table` to make this component more reusable.

akash1810: On second look, it looks like this is a pattern we've already established. Ignore this comment then.

PR #213, opened 2020-05-18
only read config from Config object

akash1810:
Keeping config co-located makes it easier to reason about.


app/config/Config.scala
```diff
@@ -85,6 +85,9 @@ object Config extends AwsInstanceTags {
 
   lazy val sessionId: String = UUID.randomUUID().toString
 
-  implicit val defaultExecutionContext =
-    scala.concurrent.ExecutionContext.Implicits.global
+  implicit val defaultExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
+
+  lazy val adminWhitelist: List[String] = getConfigStringList("application.admin.whitelist").right.getOrElse(List.empty)
+
+  lazy val capiKey: String = getConfigStringOrFail("capi.key")
```
akash1810, line 92: this is the main change

akash1810:
Whilst spiking the Play upgrade, @mbarton @jonathonherbert and I identified our configuration was a bit complicated. The Play upgrade might be a long running task, so get this out in the mean time as a general improvement.

PR #215, opened 2020-05-19
Revert "Splits preview link from open in column links"

SHession:
Reverts guardian/workflow-frontend#212. This PR has caused the preview option to no longer appear to most users, reverting for now.


akash1810:
Is it worth making the preview link column one of the core columns that cannot be hidden?

SHession:
> Is it worth making the preview link column one of the core columns that cannot be hidden?

Yeah that's a good idea, I'll see if I can find out if that would be better

paperboyo:
> Is it worth making the preview link column one of the core columns that cannot be hidden?

Or hijack prefs for Open in… for it? Got some shitty mockups *for discussion only* (to also tighten the spacing as `Preview` is too wordy):

mockup

PR #216, opened 2020-05-19
bump aws and simplify credentials provider

akash1810:
Just keeping things up to date.

Using a simpler CredentialsProvider as we don't need all the items in the current chain.

### DIFF
--- a/common-lib/src/main/scala/com/gu/workflow/util/AWS.scala
+++ b/common-lib/src/main/scala/com/gu/workflow/util/AWS.scala
@@ -1,31 +1,39 @@
 package com.gu.workflow.util
 
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
-import com.amazonaws.auth.{AWSCredentialsProviderChain, DefaultAWSCredentialsProviderChain, EnvironmentVariableCredentialsProvider, InstanceProfileCredentialsProvider}
+import com.amazonaws.auth.{AWSCredentialsProviderChain, InstanceProfileCredentialsProvider}
 import com.amazonaws.regions.{Region, Regions}
-import com.amazonaws.services.cloudwatch.{AmazonCloudWatchAsyncClient, AmazonCloudWatchAsyncClientBuilder}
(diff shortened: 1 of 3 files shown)


common-lib/src/main/scala/com/gu/workflow/util/AWS.scala
```diff
@@ -1,31 +1,39 @@
...
   lazy val region: Region = Region getRegion Regions.EU_WEST_1
 
-  lazy val EC2Client = AmazonEC2ClientBuilder.standard.withRegion(region.getName).build
-  lazy val CloudWatch = AmazonCloudWatchAsyncClientBuilder.standard.withRegion(region.getName).build
-  lazy val DynamoDb = AmazonDynamoDBClientBuilder.standard.withRegion(region.getName).withCredentials(
-    new AWSCredentialsProviderChain(
-      new EnvironmentVariableCredentialsProvider(),
-      InstanceProfileCredentialsProvider.getInstance(),
-      new ProfileCredentialsProvider("workflow"),
-      new DefaultAWSCredentialsProviderChain
```
akash1810, line 26: We're [using an InstanceProfile](https://github.com/guardian/editorial-tools-platform/blob/master/cloudformation/workflow-account/workflow/workflow-frontend.yaml#L326), so provider chain [can become a lot simpler](https://github.com/guardian/workflow-frontend/pull/216/files#diff-2b19a54750b96913359e3934504c2b0dR15).

AWare: What about running locally? Shouldn't that still use the profile provider?

akash1810: > What about running locally? Shouldn't that still use the profile provider?

Yep - https://github.com/guardian/workflow-frontend/pull/216/files#diff-2b19a54750b96913359e3934504c2b0dR16

akash1810: GH doesn't render that link very obviously. An image of the diff maybe?

image

AWare: ohh, I see. I thought that AWS was _AWS_

PR #218, opened 2020-05-20
update title attribute of commissioned length

akash1810:
Define which word count field we're comparing to for clarity.

title


paperboyo:
A tad wordy, but super useful! I would love everything to have a tooltip. A tree, a door, a sky…

PR #219, opened 2020-05-21
remove println statements

akash1810:
They're making the console a little noisy and logs should be used instead.

img

### DIFF
--- a/project/PlayArtifact.scala
+++ b/project/PlayArtifact.scala
@@ -41,9 +41,6 @@ object PlayArtifact extends Plugin {
       }
       IO.zip(playArtifactResources.value, distFile)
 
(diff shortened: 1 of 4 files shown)


project/PlayArtifact.scala
```diff
@@ -41,9 +41,6 @@ object PlayArtifact extends Plugin {
       }
       IO.zip(playArtifactResources.value, distFile)
 
-      // Tells TeamCity to publish the artifact => leave this println in here
```
AWare, line 44: this may be a can of worms, but isn't this covered by the riffraff plugin?

akash1810: It is in later versions IIUC. Given that it's commented out it's not getting executed so should have no impact.

PR #220, opened 2020-05-21
Sticky row headings

akash1810:
# Problem
We display web, print and commissioned word counts next to each other. As we scroll through the table, the column headers disappear. This means you have to build up positional definitions as they are just three numbers next to each other.

This can be confusing.

# Solution
This PR keeps the column headers in view when scrolling meaning we no longer need to maintain positional definitions.

This creates a simpler experience.

# Screenshots
## Before
before

## After
after

# Contentious changes
- The column compactor toggle has moved across a column, to `priority` (inline comments added)
  - `priority` column is now **always** shown as a result (only 6 out of 2340 people in PROD have turned off this column). This consumes either 26px or 29px depending if in compact mode or not.
- The column selector is now a sibling of the `table` rather than a child. This was the easiest way to get it displaying correctly across browsers.
- The table header is now always `48px`. Previously, the header was `37px` or `48px` depending on whether or not you had the `Web words` or `Print words` column shown as these column titles spanned across two lines. Fixing the height makes calculating the `top` offset for the group headings simpler.

# How to test
- Checkout branch
- Scroll
- Configure columns
- Scroll

You should witness a sticky table header and grouping headers across browsers, regardless of your chosen column configuration.

### DIFF
--- a/public/components/content-list/_content-list.scss
+++ b/public/components/content-list/_content-list.scss
@@ -10,6 +10,7 @@ $content-list-icon-size: 14px;
 $compact-display-td-padding: 2px 5px;
 
 $textualIconSize: 22px;
+$tableHeaderRowHeight: 48px;
 
 // Global padding for cells
 %content-list__field-padding {
@@ -29,18 +30,25 @@ $textualIconSize: 22px;
     width: 100%;
     border-bottom: 5px solid $c-grey-100;
 
-    position: relative;
+    thead {
+        th {
+            z-index: 3; // always keep table header on top, even on top of the column selector
+            background-color: $c-white;
+        }
+        tr {
+            height: $tableHeaderRowHeight;
+        }
+    }
 
     &--sticky-row th {
         position: sticky;
         top: 0;
-        background-color: $c-white;
-        z-index: 1;
+        background-color: $c-bluegrey;
 
         &.content-list__group-heading {
-            $tableHeaderRowHeight: 37px; // ideally this would be calculated dynamically
-            top: $tableHeaderRowHeight; // keep `status` row headings below the column header row
-            background-color: $c-bluegrey;
+            // Chrome has a gap between the table header and a group heading row
+            $strangeChromeGap: 1px;
+            top: $tableHeaderRowHeight - $strangeChromeGap;
         }
     }
     // Table header
@@ -285,7 +293,7 @@ $textualIconSize: 22px;
 
 .column-configurator {
     position: absolute;
-    top: 34px;
+    top: $tableHeaderRowHeight;
     right: 6px;
     width: 180px;
     z-index: 2;
--- a/public/components/content-list/_content-list.scss
+++ b/public/components/content-list/_content-list.scss
@@ -43,7 +43,7 @@ google-auth-banner .alert {
 
     thead {
         th {
-            z-index: 3; // always keep table header on top, even on top of the column selector
+            z-index: 3;
             background-color: $c-white;
         }
         tr {
@@ -58,6 +58,7 @@ google-auth-banner .alert {
 
         &.content-list__group-heading {
             top: $tableHeaderRowHeight;
+            z-index: 2;
         }
     }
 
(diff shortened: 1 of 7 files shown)


jonathonherbert:
The scrolling works well, but it looks like the column menu is obscured by the table header –
Screenshot 2020-05-22 at 09 19 42
Bumping the `.column-configurator` z-index by 1 seems to solve the issue.

akash1810:
> but it looks like the column menu is obscured by the table header

Hmm, can't reproduce. What browser and version are you using?

jonathonherbert:
Chrome, Version 81.0.4044.138 (Official Build) (64-bit) – bit of a strange one. I've just noticed that the alignment of the popover is off on smaller widths, too, but it sounds like you're going down the modal route anyway, so this may no longer be relevant. Happy to pair to help reproduce, here's a gif to better show how I got there –

workflow-trubs

akash1810:
> I've just noticed that the alignment of the popover is off on smaller widths, too

The popover is now fixed. What was the width of your viewport?

akash1810:
> The scrolling works well, but it looks like the column menu is obscured by the table header

This is because I hadn't considered the blue Google auth alert banner in the positioning. https://github.com/guardian/workflow-frontend/pull/220/commits/a01ecc7ac6db0da48f06700334f5b7207a760f0d fixes that. Great spot!

public/components/content-list/_content-list.scss
```diff
@@ -295,17 +341,30 @@ $textualIconSize: 22px;
...
+            cursor: pointer;
+            width: 100%;
+
+            input {
+                width: auto;
+            }
+
+            &:hover {
+                background-color: $c-grey-200;
+            }
```
akash1810, line 367: This creates a more accessible UX.

**Before**
menu


**After**
menu

paperboyo: This one is great! One tiny nitpick would be for the label text not tot be selectable, as visible in GIF in “Assignee”.

paperboyo:
1. Columns picker opens only half-height for me requiring scrolling even if all options would comfortably fit:

image

2. Column compactor is gone:

image

vs [PROD]:

image

akash1810:
> 2. Column compactor is gone:

😱 What browser?

public/components/content-list/_content-list.scss
akash1810: > the label text not tot be selectable

Fixed in https://github.com/guardian/workflow-frontend/pull/220/commits/65825e4fa23a97298ee0baa15e1f18f6bbf777a0.

akash1810:
> 1. Columns picker opens only half-height for me requiring scrolling even if all options would comfortably fit:

Ah, good point. The max height is [now](https://github.com/guardian/workflow-frontend/commit/65825e4fa23a97298ee0baa15e1f18f6bbf777a0) relative to viewport size, rather than frozen at `400px`.

paperboyo:
> 😱 What browser?

Firefox 76.0.1, Chrome 81.0.4044.138 & 83.0.4103.61 on OSX 10.14.6 (I’m in an expanded state)

> The max height is [now](https://github.com/guardian/workflow-frontend/commit/65825e4fa23a97298ee0baa15e1f18f6bbf777a0) relative to viewport size

👏

akash1810:
> Firefox 76.0.1, Chrome 81.0.4044.138 & 83.0.4103.61 on OSX 10.14.6

Same here 🤔 but can reproduce. Can you share your browser window width and the column configuration you've got?

paperboyo:
> Can you share your browser window width and the column configuration you've got?

Sorry, in the middle of a roadtrip back… Priority field seems to be involved, also every now and then I am still able to scroll the whole table horizontally:

presence

akash1810:
> Priority field seems to be involved

Fixed in https://github.com/guardian/workflow-frontend/pull/220/commits/bddc25cdd5c67d6b6248261dc622e3cc4b93589f by not allowing the `priority` column to be shown - only 6 out of 2340 people have it turned off in PROD, so impact isn't huge.

AWare:
Why does priority being hidden hide the compactor?

akash1810:
> Why does priority being hidden hide the compactor?

The column compactor toggle has moved across a column, to priority (inline comments added).

TL;DR the compactor was absolute positioned, which broke when switching to sticky. Moving it across provided a wider column and meant absolute positioning could be removed.

PR #221, opened 2020-05-22
update presence styling in drawer

akash1810:
# Problem
In the drawer, we're displaying the full name, rather than the initials, for the presence indicator. The background currently doesn't span the entire text, making it unreadable.

# Solution
Allow the background to span the full name to make it more readable.

Presence indicators in the table remain unchanged (they are still a circle with initials).

# Screenshots
## Before
image

## After
image


jonathonherbert:
Looks good on CODE!

PR #222, opened 2020-05-22
Use CODE Tools in DEV

akash1810:
This makes development simpler as you don't need to start N other services (Preferences, Presence, Composer, etc).

### DIFF
--- a/app/config/Config.scala
+++ b/app/config/Config.scala
@@ -26,34 +26,12 @@ object Config extends CommonConfig with AwsInstanceTags {
   lazy val host: String = s"https://workflow.$domain"
   Logger.info(s"Host is: $host")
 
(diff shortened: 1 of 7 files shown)


sihil:
I generally disagree with this strategy as it makes it harder to work on any integration where you're testing dev versions of various services. It's obviously too hard to work locally but you generally don't need to start other services unless you are working on them and if you are working on them then you'd want the dev version running.

If the consensus is that this should go in then by all means do so but this is a vote against.

akash1810:
> as it makes it harder to work on any integration where you're testing dev versions of various services

Maybe employing a similar strategy as Composer is better? That is, push all the endpoints to config to make it simpler to change if needed?

akash1810:
There's general consensus that making it easier to run everything locally is preferred to this. Closing.

PR #226, opened 2020-05-26
Display print location information

aug24:
Print location information column is added to the workflow table.

Normal:
image
Compact:
image


common-lib/src/main/scala/com/gu/workflow/api/StubAPI.scala
```diff
@@ -0,0 +1,283 @@
...
+                                           externalData: Option[ExternalData]
+                                         ): Future[Option[String]] = {
+    externalData
+      .flatMap(e => Some(e.actualPublicationId, e.actualBookId, e.actualBookSectionId)) match {
+      case Some((Some(pId), Some(bId), Some(bsId))) => getPrintLocationDescriptionFromIds(contentAPI, pId, bId, bsId)
+      case _ => Future.successful(None)
+    }}
+
+  private def getPrintLocationDescriptionFromIds(contentAPI: ContentAPI, pId: Long, bId: Long, bsId: Long): Future[Option[String]] =
+    memoize(60.minutes) {
```
jonathonherbert, line 63: Discussed over chat but for posterity – memoizing at this level will produce a new cache key for every new set of arguments, even if some of those arguments have been seen within the cache window. For example, calls to the parent with pId, bId and bsId (1,2,3), (1,2,4) and (1,2,5) will all result in fresh invocations, although ids 1 and 2 are shared.

It's probably better to cache within `contentAPI.getTagInternalName`, which also has the benefit of hiding details of the cache implementation from consuming code.

aug24: Done.

akash1810:
At first glance, this is doing more than reverting a commit - can we add more detail to the description please?

aug24:
> At first glance, this is doing more than reverting a commit - can we add more detail to the description please?

Very good point. Done.

.java-version
```diff
@@ -1 +1 @@
-1.8.0.242
+1.8.0.202
```
akash1810, line 1: Why so low?

app/controllers/Api.scala
```diff
@@ -319,3 +321,21 @@ object Api extends Controller with PanDomainAuthActions with SharedSecretAuth {
         block(req)
   }
 }
+
+//      def getTag(publicationId: Option[Int],
```
akash1810, line 325: Why are these commented out?

public/components/content-list-item/_content-list-item.scss
```diff
@@ -541,3 +544,10 @@
         display: inline-block;
     }
 }
+
+.planned-print-location {
+    color: grey;
```
akash1810, line 549: Could we re-use something from the [palette](https://github.com/guardian/workflow-frontend/blob/master/public/layouts/global/_colours.scss)?

akash1810:
Could we add some screenshots of the table in normal and compact mode?

app/controllers/Api.scala
aug24: Carelessness?

aug24:
> Not sure I understand/follow why we're moving from `CommonAPI` and `PrototypeAPI` to `StubAPI` 

Because _both_ those API class actually dealt with serving Stub information.

> and why we're caching?

For speed, mostly.

public/components/content-list-item/_content-list-item.scss
aug24: Done.

.java-version
aug24: Changed to 232.

jonathonherbert:
A bit more detail on the cacheing @akash1810 – because we need to get tags on each stub request (which happens often), because we expect content to have similar tags, and we don't expect that tags will change very often, cacheing tag information seemed like a good way of reducing the load on CAPI and improving the speed at which we serve requests.

It wouldn't be appropriate if we expected tag information to be extremely up to date.

jonathonherbert:
Along those lines, @aug24 – what happens when CAPI is down? Do we have a reasonable timeout on the requests we make?

PR #229, opened 2020-05-28
Upgrade node to 12

AWare:
- adds metals and brewfile to gitignore
- regenerate the yarn lockfile because of an fsevents discrepancy 
- update the js tests so they pass again (I'm not convinced they have since 2017)


AWare:
(I checked teamcity and there's no explicit dependency on version there)

PR #231, opened 2020-05-28
Upgrade babel and webpack

AWare:
This updates babel and webpack because they're both quite old and this unblocks moving the frontend architecture forwards easier.

### DIFF
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -15,7 +15,7 @@
             padding-top: 72px;
         }
     </style>
-    <link rel="stylesheet" type="text/css" href="@routes.Assets.versioned("build/main.css")"/>
+    <link rel="stylesheet" type="text/css" href="@routes.Assets.versioned("build/app.css")"/>
 </head>
 
 <body>
--- a/conf/webpack.conf.js
+++ b/conf/webpack.conf.js
@@ -1,10 +1,14 @@
 /* global module:false, __dirname:false */
 
-var path = require('path');
-var webpack = require('webpack');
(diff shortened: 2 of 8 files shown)


conf/webpack.conf.js
```diff
@@ -14,36 +18,39 @@ module.exports = {
...
               exclude: '/app/',
               loader: 'html-loader'
             },
-            {
-              test: /\.json$/,
-              exclude: /node_modules/,
-              use: 'json-loader'
-            },
             {
                 test: /\.scss$/,
```
jonathonherbert, line 35: By changing this test to /\.s?css$/, can we avoid two CSS rules?

app/views/main.scala.html
```diff
@@ -15,7 +15,7 @@
             padding-top: 72px;
         }
     </style>
-    <link rel="stylesheet" type="text/css" href="@routes.Assets.versioned("build/main.css")"/>
+    <link rel="stylesheet" type="text/css" href="@routes.Assets.versioned("build/app.css")"/>
```
jonathonherbert, line 18: Why the change?

jonathonherbert:
Works fine locally 🎉 , but the production configuration breaks JS with an 'unknown provider' error that looks like a DI minification problem.

conf/webpack.conf.js
AWare: Yeah, let's scrap that extra ruke.

app/views/main.scala.html
AWare: This splits out the main and admin css into different files, whereas previously it did not.

PR #234, opened 2020-05-30
Normalise font size for Expanded view

paperboyo:
Currently, in Expanded view, Publication location and both Wordcount fields use larger font. I don’t understand why they do not inherit font sizes like other fields , but overrides like these already exist for Compact view and I just wanted to see if it’s gonna work. Ideally, we wouldn’t have these special cases and font sizes would be consistent by themselves, though, I suppose.

This is a draft to deploy to CODE and see. And for someone, hopefully, to say how this should be done correctly. 

- [x] tested on CODE

This achieves the desired effect:
from:
image
to:
image

Unfortunately, all overridden text sizes are in pixels (`11px`), while the correctly set ones are in rems (`1.1rem`). This may cause differences when setting font size via browser settings or something…

### DIFF
--- a/public/components/content-list-item/_content-list-item.scss
+++ b/public/components/content-list-item/_content-list-item.scss
@@ -164,15 +164,21 @@
         &--ophan,
         &--created,
         &--incopy,
-        &--wordcount,
-        &--printwordcount,
+        &--wordcount {
+            @extend %fs-data-1;
+        }
+
+        &--printwordcount {
+            @extend %fs-data-1;
+        }
+
         &--commissionedLength,
         &--needsLegal,
         &--optimisedForWeb,
         &--sensitive,
         &--legally-sensitive,
         &--publicationlocation {
-            @extend %fs-data-3;
+            @extend %fs-data-1;
             @extend %content-list__field-padding;
         }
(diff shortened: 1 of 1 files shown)


public/components/content-list-item/_content-list-item.scss
```diff
@@ -164,15 +164,21 @@
...
-        &--wordcount,
-        &--printwordcount,
+        &--wordcount {
+            @extend %fs-data-1;
+        }
+
+        &--printwordcount {
+            @extend %fs-data-1;
+        }
+
```
akash1810, line 174: This can be simplified to the following as it looks like all the columns now share the same `@extend %fs-data-1` styles?

```suggestion
        &--wordcount,
        &--printwordcount
```

public/components/content-list-item/_content-list-item.scss
```diff
@@ -165,14 +165,14 @@
         &--created,
         &--incopy,
         &--wordcount,
-        &--printwordcount,
+        &--printwordcount
```
akash1810, line 168:
```suggestion
        &--printwordcount,
```

PR #236, opened 2020-06-01
Improve "create new" modal behaviour

aug24:
## What does this change?

Currently, pressing enter cancels the modal.  This change improves behaviour such that pressing enter on the modal does the default (displayed button) action.

No behavioural changes other than the improved modal behaviour.

### DIFF
--- a/public/components/stub-modal/stub-modal.html
+++ b/public/components/stub-modal/stub-modal.html
@@ -114,6 +114,9 @@
     </div>
 
     <div class="modal-footer" ng-if="mode !== 'import'" ng-hide="actionInProgress || (actionSuccess && composerUrl)">
+        <button type="submit" class="btn btn-default" id="testing-do-generic-submit" ng-click="submit(stubForm)" ng-show="false">
+            Submit
+        </button>
         <div class="pull-left">
             <button type="button" class="btn btn-default pull-left" data-dismiss="modal" ng-click="delete()" ng-if="stub.id">Delete</button>
             <button type="button" class="btn btn-default" data-dismiss="modal" ng-click="cancel()">Cancel</button>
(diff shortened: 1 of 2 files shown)


public/components/stub-modal/stub-modal.html
```diff
@@ -114,25 +114,28 @@ <h4 class="modal-title">{{ modalTitle }}</h4>
...
+        <button type="submit" class="btn btn-default" id="testing-do-generic-submit" ng-click="submit(stubForm)" ng-show="false">
+            Submit
+        </button>
         <div class="pull-left">
             <button type="button" class="btn btn-default pull-left" data-dismiss="modal" ng-click="delete()" ng-if="stub.id">Delete</button>
             <button type="button" class="btn btn-default" data-dismiss="modal" ng-click="cancel()">Cancel</button>
         </div>
         <div class="pull-right">
-            <button type="submit" class="btn btn-default" id="testing-create-stub" ng-click="ok(addToComposer=false,addToAtomEditor=false)" ng-disabled="disabled || stubForm.$invalid " ng-show="stub.status === 'Stub'">
+            <button type="button" class="btn btn-default" id="testing-create-stub" ng-click="submit(stubForm)" ng-disabled="disabled || stubForm.$invalid " ng-show="stub.status === 'Stub'">
```
jonathonherbert, line 125: Minor: if only one of these buttons is displayed at the same time, could we remove the click handlers, make them type `submit`, and bind the submit function to the form with `ng-submit={submit}` once?

jonathonherbert: (looking at the code below, there's a chance this might not play nicely with `actionSuccess`)

PR #237, opened 2020-06-02
refactor print location decoration for simplicity

akash1810:
## What does this change?
The current logic is a bit confusing as it's moving a type of `Future[(Option[String], Option[String])]` around and consequently using positional references, which is hard to follow.

Refactor to:
- push print location decoration into its own class
- introduce DecoratedPrintLocation in place of a tuple of optional strings

Easier to understand the print location logic.

## Images
n/a

### DIFF
--- a/common-lib/src/main/scala/com/gu/workflow/api/StubAPI.scala
+++ b/common-lib/src/main/scala/com/gu/workflow/api/StubAPI.scala
@@ -43,7 +43,10 @@ object StubAPI {
       res <- ApiResponseFt.Async.Right(getRequest(s"content/$composerId"))
       json <- parseBody(res.body)
       maybeStub <- extractDataResponseOpt[Stub](json)
-      maybeDecoratedStub <- ApiResponseFt.Async.Right(Future.traverse(maybeStub)(stubDecorator.withPrintLocationDescriptions))
+      maybeDecoratedStub <- ApiResponseFt.Async.Right(maybeStub match {
+        case Some(stub) => stubDecorator.withPrintLocationDescriptions(stub).map(Some(_))
+        case _ => Future.successful(None)
+      })
     } yield maybeDecoratedStub
   }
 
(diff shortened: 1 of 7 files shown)


common-lib/src/main/scala/com/gu/workflow/api/StubAPI.scala
```diff
@@ -36,60 +38,17 @@ object StubAPI {
...
-              shortActualPrintLocationDescription = actualPrintLocationDescription._2)
-          )
-        )
-        .copy(
-          longPlannedPrintLocationDescription = plannedPrintLocationDescription._1,
-          shortPlannedPrintLocationDescription = plannedPrintLocationDescription._2
-        )
-      )
+      maybeDecoratedStub <- ApiResponseFt.Async.Right(maybeStub match {
+        case Some(stub) => stubDecorator.withPrintLocationDescriptions(stub).map(Some(_))
```
akash1810, line 47: `.map(Some(_))` is necessary to satisfy the type. It's not great and help wanted to improve it!

aug24:
```suggestion
ApiResponseFt.Async.Right(maybeStub.map(stub => stubDecorator.withPrintLocationDescriptions(stub)))
```

common-lib/src/main/scala/com/gu/workflow/api/StubAPI.scala
```diff
@@ -36,60 +38,17 @@ object StubAPI {
...
-              longActualPrintLocationDescription = actualPrintLocationDescription._1,
-              shortActualPrintLocationDescription = actualPrintLocationDescription._2)
-          )
-        )
-        .copy(
-          longPlannedPrintLocationDescription = plannedPrintLocationDescription._1,
-          shortPlannedPrintLocationDescription = plannedPrintLocationDescription._2
-        )
-      )
+      maybeDecoratedStub <- ApiResponseFt.Async.Right(maybeStub match {
```
aug24, line 46:
```suggestion
      maybeDecoratedStub <- 
```

common-lib/src/main/scala/com/gu/workflow/api/StubAPI.scala
akash1810: 😢  this doesn't satisfy the types.

```scala
ApiResponseFt.Async.Right(maybeStub.map(stub => stubDecorator.withPrintLocationDescriptions(stub)))
```

yields an `Option[Future[Stub]]`, but a `Future[Option[Stub]]` is needed.

PR #238, opened 2020-06-02
Work to allow planned print locations to be edited

AWare:
## What does this change?

This adds the functionality to edit planned print info fields.

Todo:

- [x] api changes
- [x] workflow datastore changes

### DIFF
--- a/conf/routes
+++ b/conf/routes
@@ -42,7 +42,11 @@ PUT            /api/stubs/:stubId/workingTitle             controllers.Api.putSt
 PUT            /api/stubs/:stubId/priority                 controllers.Api.putStubPriority(stubId: Long)
 PUT            /api/stubs/:stubId/trashed                  controllers.Api.putStubTrashed(stubId: Long)
 PUT            /api/stubs/:stubId/status                   controllers.Api.putStubStatus(stubId: Long)
-PUT            /api/stubs/:stubId/commissionedLength       controllers.Api.putStubCommissionedLength(stubId: Long)
+PUT            /api/stubs/:stubId/plannedPublicationId     controllers.Api.putStubPlannedPublicationId(stubId: Long)
+PUT            /api/stubs/:stubId/plannedBookId            controllers.Api.putStubPlannedBookId(stubId: Long)
+PUT            /api/stubs/:stubId/plannedBookSectionId     controllers.Api.putStubPlannedBookSectionId(stubId: Long)
+PUT            /api/stubs/:stubId/plannedNewspaperPageNumber          controllers.Api.putStubPlannedNewspaperPageNumber(stubId: Long)
+# PUT            /api/stubs/:stubId/plannedNewspaperPublicationDate     controllers.Api.putStubPlannedNewspaperPublicationDate(stubId: Long)
 DELETE         /api/stubs/:stubId                          controllers.Api.deleteStub(stubId: Long)
 
 GET            /api/statuses                               controllers.Api.statusus
--- a/conf/routes
+++ b/conf/routes
@@ -46,7 +46,7 @@ PUT            /api/stubs/:stubId/plannedPublicationId     controllers.Api.putSt
 PUT            /api/stubs/:stubId/plannedBookId            controllers.Api.putStubPlannedBookId(stubId: Long)
 PUT            /api/stubs/:stubId/plannedBookSectionId     controllers.Api.putStubPlannedBookSectionId(stubId: Long)
 PUT            /api/stubs/:stubId/plannedNewspaperPageNumber          controllers.Api.putStubPlannedNewspaperPageNumber(stubId: Long)
-# PUT            /api/stubs/:stubId/plannedNewspaperPublicationDate     controllers.Api.putStubPlannedNewspaperPublicationDate(stubId: Long)
+PUT            /api/stubs/:stubId/plannedNewspaperPublicationDate     controllers.Api.putStubPlannedNewspaperPublicationDate(stubId: Long)
 DELETE         /api/stubs/:stubId                          controllers.Api.deleteStub(stubId: Long)
 
 GET            /api/statuses                               controllers.Api.statusus
--- a/public/components/content-list-drawer/content-list-drawer.js
+++ b/public/components/content-list-drawer/content-list-drawer.js
@@ -17,8 +17,7 @@ var SETTING_OPEN_SECTION = 'openSection';
  * @param contentService
  * @param prodOfficeService
  */
-export function wfContentListDrawer($rootScope, config, $timeout, $window, contentService, prodOfficeService, wfGoogleApiService, wfCapiContentService, wfCapiAtomService, wfAtomService, wfSettingsService, wfComposerService) {
-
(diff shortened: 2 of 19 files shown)


conf/routes
```diff
@@ -44,7 +44,11 @@ PUT            /api/stubs/:stubId/workingTitle             controllers.Api.putSt
 PUT            /api/stubs/:stubId/priority                 controllers.Api.putStubPriority(stubId: Long)
 PUT            /api/stubs/:stubId/trashed                  controllers.Api.putStubTrashed(stubId: Long)
 PUT            /api/stubs/:stubId/status                   controllers.Api.putStubStatus(stubId: Long)
-PUT            /api/stubs/:stubId/commissionedLength       controllers.Api.putStubCommissionedLength(stubId: Long)
+PUT            /api/stubs/:stubId/plannedPublicationId     controllers.Api.putStubPlannedPublicationId(stubId: Long)
+PUT            /api/stubs/:stubId/plannedBookId            controllers.Api.putStubPlannedBookId(stubId: Long)
+PUT            /api/stubs/:stubId/plannedBookSectionId     controllers.Api.putStubPlannedBookSectionId(stubId: Long)
+PUT            /api/stubs/:stubId/plannedNewspaperPageNumber          controllers.Api.putStubPlannedNewspaperPageNumber(stubId: Long)
+# PUT            /api/stubs/:stubId/plannedNewspaperPublicationDate     controllers.Api.putStubPlannedNewspaperPublicationDate(stubId: Long)
```
akash1810, line 51: Should we have a single endpoint for planned newspaper data?

The current `PUT` routes in this file appear to be for fields that are independent of each other and the data that constitute planned paper information is tightly coupled.

Having a single endpoint would allow for stronger input validation, I think, too.

AWare: There aren't any `PUT` routes in this project which combine multiple fields. I don't think this warrants a second type of `PUT` endpoint, considering that the existing ones are so similar to each other in implementation.

AWare: Also, if these three fields are so bound- then they should probably be modeled as a case clas..

AWare: This probably should have been part of https://github.com/guardian/workflow-frontend/pull/226 can this model change be raised as a new card in trello?

AWare:
image
image

public/components/content-list-drawer/content-list-drawer.js
```diff
@@ -379,6 +398,43 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
...
             });
+
+            $scope.setBookSections = function() {
+                var results = wfTagApiService.searchTags($scope.bookSectionQuery, 'newspaper book section');
+                results.then(function (response) {
+                const tags = (response && response.data) || []
+                    return tags.map(function (tag) {
+                        return { item: tag.data };
+                    });
+                }).then(function(candidateBookSections) {
```
jonathonherbert, line 409: Any need for this second `then`? The value returned from the previous block doesn't look like another promise.

public/lib/tag-api-service.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  function getTag(id) {
+    return $http({
+      method: "GET",
+      url: `${url}/${id}`,
+    }).then(extractHyper);
+  }
+
+  function getHyperTag(hyperTag) {
```
jonathonherbert, line 33: It'd be great to document this with an example of what it's likely to process, what it returns, and why – ATM the reader does not know what a `hyperTag` is 😁

public/components/directives/ui-filter-list.js
```diff
@@ -0,0 +1,147 @@
+import angular from 'angular';
+import _ from 'lodash';
+import getKeyCodes from 'keycode'
+ 
+export function uiFilterList($q, $window) {
```
jonathonherbert, line 5: Presume this is a copy from Composer?

jonathonherbert: (if so, and if trivial, the spec could come too 😁)

public/components/content-list-drawer/content-list-drawer.js
AWare: At 440 lines, you can forgive me for my overenthusiasm there. :(

jonathonherbert: Oh, for sure, no criticism – just pedantic linting 😁

public/components/directives/ui-filter-list.js
AWare: (We don't run the tests in this project)

jonathonherbert: haha, naw 😅

AWare: Having a good look at this test, it uses a fair number of composer helpers.

jennygrahamjones:
Can we add some tests? 

Also, is it possible to manually test this yet, or does that depend on it being unblocked by work around saving planned print info?

AWare:
This will run in code, but not save until the other work is done. 

Because of the nightmare this was, and the amount of workflow work desired for the next quarter- I'm going to spike the drawer in react.

AWare:
From @paperboyo 
1. Highlights should be blue
2. Duplication in firefox on dropdown
3. Pub date should not show hours.
4. Validate page number is number
5. Fix the heights so things don't move on edit.

AWare:
2. 
 
Kapture 2020-06-21 at 10 49 21
Kapture 2020-06-21 at 10 51 33

Not sure what to make of this one, it's also present in composer. It's down to the fake dropdown maintain separate keyboard and mouse states. Personally, I'd like to redo the whole panel in React and just use someone else's dropdown component.

paperboyo:
> Not sure what to make of this one

I know! Ignore (then). :-)

AWare:
Did the rest though.

paperboyo:
6. Clicking on editing a page no. correctly immediately focuses field allowing entry straight away, not so for Booksection and Date (cursor not in field)
7. Clearing values should be possible

jennygrahamjones:
@AWare there are some TODOs in the PR description. Are they still outstanding, or work for another PR? Also, what's the best way to test this manually?

AWare:
I can deploy them both to code if you'd like. 
Otherwise, deploy the equivalent workflow to code and use the ssh tunnel.

jennygrahamjones:
Looks good, in terms of being able to update and store the details.

One question about the 'Planned Publication Date'. When I select the date, that value is displayed for a moment, before the date prior to that one replaces it. See the GIF below for an example: I select 1st July, but the date changes to 30th June. Is this intended?

AWare:
That whole pane is weird custom saving code, I'm not sure why it's flashing up as the wrong date.

jennygrahamjones:
> That whole pane is weird custom saving code, I'm not sure why it's flashing up as the wrong date.

It appears to be either displaying or saving the wrong day, then? If I pick July 1st, it will change to June 30th and after saving/reloading, June 30th is still the date that's there.

AWare:
Time zones strike again!

AWare:
The date seems to make a timezone related oscillation at every possible step.

AWare:
Ok, this is ready again

PR #242, opened 2020-06-04
React!

AWare:
## What does this change?
Everything.
In the years regained by not having variables auto recased, using dependencies without having to magically inject them. 
## Images
image

## Seriously

This is a spike to show that we can develop react things inside angular fairly reasonably. It was nightmarish implementing https://github.com/guardian/workflow-frontend/pull/238/files in angular, had this been react I wouldn't have had problems such as:
- isolating the typeahead dropdown / finding a new one
- not being able to use the pre-existing `wf-editable` component for the date and dropdown
- having to hand wire savers and reverts for the new fields
- things-getting-renamed-from-this ToSomeThingThatLooksLikeThis so you can't grep for them 

(nb this uses this as a base https://github.com/guardian/workflow-frontend/pull/231 )

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["@babel/preset-env"]
+  "presets": ["@babel/preset-env","@babel/preset-react"]
 }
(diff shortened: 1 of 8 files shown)


SHession (dismissed):
This is really cool! Tested it locally and it all works correctly. I've also looked through the code and LGTM! What's the plan for this moving forward?

akash1810:
Closed as stale, and quite old, which are signals that this is not important/needed/

PR #243, opened 2020-06-04
Fixes filtering midnight dates

SHession:
## What does this change?
This change minuses a second from dates being used to filter content. This prevents dates set to midnight being filtered out erroneously. This was necessary as the database query treats the `from` values as a `greater than` and not `great than or equal to` comparison.

The deadline filters should now not filter out content set to midnight. In particular content set to a deadline of midnight on Saturday, should appear under the weekend deadline filter. This should also fix filtering created dates, although they are unlikely to be set to midnight.


public/lib/date-service.js
```diff
@@ -136,6 +136,20 @@ angular.module('wfDateService', ['wfLocationService'])
             return this.createRange(dayStart, dayStart.clone().add(1, 'days'));
         }
 
+        /**
+         * Minus a second from a "from" date to avoid midnight dates being filtered out
```
jennygrahamjones, line 140: Could we amend/expand this comment slightly? This would be applicable in any situation where we want to filter from a specific time onward, e.g. if we had a filter for 'This afternoon' that had to include everything with a deadline from 12pm onwards. Maybe we could add a little context about how the query works and why we need this work around.

SHession: You're right, that comment is a bit sparse. I will add some context to it.

PR #253, opened 2020-06-18
Sort workflow columns

jonathonherbert:
## What does this change?

This PR makes some workflow columns sortable on click of the column header.

sort-workflow-cols

The list of sortable columns can grow as what is represented in the template vs the model is reconciled. The list of columns made sortable by this PR should be clear when reviewing `column-defaults.js`.

Click on the column header of any column referenced as sortable in `column-defaults.js`. You should see an emoji displaying the sort direction on the column. The content should be sorted accordingly.

Toggling through the states should work as expected – asc, desc, none, repeat. Toggling different columns should reset the sort order and add an ascending sort to the current column, removing the previous sort.

## Notes

- Sorting triggers the table's 'new content' highlight UX. This might feel misleading, although it could be difficult to work around – please let me know!
- At the moment, we sort ignoring lowercase, but there's no allowance for natural sort, dates etc. We will probably need to improve this to avoid misleading results in columns that require those kinds of sorting approaches.
- We'll need to model the print location data as a single, sortable value to be able to sort by that value, which I don't think is currently the case. This work probably belongs in another PR – `content-list-item.js` contains the `ContentItemView` class I think we'd need to alter. This should also allow us to simplify the `publicationLocation.html` template, too, by moving the display logic into the model (or – even better – onto the server?)
- This would be an ideal candidate for an integration test 😅 – AFAIK we don't have a framework at the moment for this project, but once one is there ... 😁

### DIFF
--- a/public/components/content-list/content-list.js
+++ b/public/components/content-list/content-list.js
@@ -111,8 +111,9 @@ function wfContentListController($rootScope, $scope, $anchorScroll, statuses, le
 
     $scope.sortColumn = undefined;
     $scope.sortDirection = undefined;
+    const defaultSortColName = 'titles';
+    const sortStates = ['asc', 'desc', undefined];
 
-    const sortStates = ['asc', 'desc', undefined]
     $scope.toggleSortState = columnName => {
       // Reset the sort order if we're toggling a new field
       const sortDirectionIndex = columnName === $scope.sortColumn
@@ -136,6 +137,12 @@ function wfContentListController($rootScope, $scope, $anchorScroll, statuses, le
 
     wfColumnService.getColumns().then((data) => {
         $scope.columns = data;
+
+        // Apply sort defaults
+        const column = $scope.columns.find(_ => _.name === defaultSortColName);
+        if (column) {
+          $scope.toggleSortState(column.sortField || column.name)
+        }
     });
 
     $scope.getColumnTitle = function(col) {
--- a/public/components/content-list/content-list.js
+++ b/public/components/content-list/content-list.js
@@ -112,7 +112,9 @@ function wfContentListController($rootScope, $scope, $anchorScroll, statuses, le
     $scope.sortColumn = undefined;
     $scope.sortDirection = undefined;
     const defaultSortColName = 'titles';
-    const sortStates = ['asc', 'desc', undefined];
+    // If we'd prefer to allow people to remove the sort state entirely,
+    // this list can be changed to ['asc', 'desc', undefined]
+    const sortStates = ['asc', 'desc'];
 
     $scope.toggleSortState = columnName => {
       // Reset the sort order if we're toggling a new field
--- a/public/components/content-list/content-list.js
+++ b/public/components/content-list/content-list.js
@@ -107,6 +107,23 @@ function wfContentListController($rootScope, $scope, $anchorScroll, statuses, le
         wfFiltersService.clearAll(false);
     };
 
+    $scope.getSortDirection = columnName => columnName === $scope.sortColumn ? $scope.sortDirection : undefined;
+
+    $scope.sortColumn = undefined;
+    $scope.sortDirection = undefined;
+
+    const sortStates = ['asc', 'desc', undefined]
+    $scope.toggleSortState = columnName => {
+      // Reset the sort order if we're toggling a new field
+      const sortDirectionIndex = columnName === $scope.sortColumn
+        ? (sortStates.indexOf($scope.sortDirection) + 1) % sortStates.length
+        : 0;
+      $scope.sortDirection = sortStates[sortDirectionIndex];
+      $scope.sortColumn = $scope.sortDirection ? columnName : undefined;
+
+      applyNewContentState();
+    };
+
     $scope.animationsEnabled = true;
 
     $rootScope.$on('getContent', () => {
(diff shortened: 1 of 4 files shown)


public/components/content-list/content-list.js
```diff
@@ -107,6 +107,23 @@ function wfContentListController($rootScope, $scope, $anchorScroll, statuses, le
...
     };
 
+    $scope.getSortDirection = columnName => columnName === $scope.sortColumn ? $scope.sortDirection : undefined;
+
+    $scope.sortColumn = undefined;
+    $scope.sortDirection = undefined;
+
+    const sortStates = ['asc', 'desc', undefined]
+    $scope.toggleSortState = columnName => {
+      // Reset the sort order if we're toggling a new field
```
jonathonherbert, line 117: Reading this back, it's a bit golf-y – lmk if we'd like to express this in a simpler (but more verbose) way.

public/lib/column-defaults.js
```diff
@@ -276,6 +297,14 @@ var columnDefaults = [{
     template: lastModifiedByTemplate,
     active: false,
     isNew: true
-}].map(col => col.labelHTML === '' ? {...col, labelHTML: '&nbsp;'} : col);
+}].map(col => {
+  if (col.labelHTML === '') {
+    return {...col, labelHTML: '&nbsp;'};
+  }
+  if (col.isSortable) {
```
akash1810, line 304: Does the `return` above mean you need a `labelHTML` to be eligible for sorting?

public/lib/column-defaults.js
```diff
@@ -34,11 +34,24 @@ import publicationLocationTemplate from "components/content-list-item/templates/
...
  *      templateUrl: string // URL for the content-list-item template for this field
+ *      isSortable: boolean // Can the column be sorted by clicking its header?
+ *      sortField: string // The field to sort on, if different from `name`. Can be an item path, e.g. `a.nested.field`
  * }
  */
 
 var templateRoot = '/assets/components/content-list-item/templates/';
 
+const createSortTemplate = (sortField, labelHTML) => `
+    <div ng-click="toggleSortState('${sortField}')" class="content-list-head__heading-sort-by">
```
akash1810, line 45: Would `content-list-head__heading--sort-by` be more BEMy?

paperboyo:
> asc, desc, none, repeat

By _repeat_ you mean _repeat sequence_?
Not sure about _none_. We should establish what is the sort when _none_ is active and stick icon in that column and get rid of _none_, I think.

Also, could we use U+25BC for Z–A and U+25B2 for A–Z? These emojis are from :hell:. Ideally, they wouldn’t move around as well like in the GIF: possible to make sure they are always in predictable one plca (like eg before the column heading)?

public/lib/column-defaults.js
jonathonherbert: It does! Coincidentally, that's true of the columns we currently sort by, but it's not what we want, thanks for spotting 👍

paperboyo:
Another thing is: Workflow is (always?) grouped by Status. How does this interact with it, is it sorting only within these Status groups still? One day (not this PR!), would be nice to be able to ungroup (better still, group by eg Day, or Booksection etc).

public/lib/column-defaults.js
jonathonherbert: As I understand it, the M of BEM is for flags – things that change state. I was considering `heading-sort-by` to be an element of the `content-list-head` block – it's not a subelement of `heading` in BEM-land. Although it _is_ a subelement of heading in the DOM. 😁

So I suppose it depends how flat you'd like the hierarchy to be – I've followed some docs in the past that suggest that [flattening elements isn't necessary](http://getbem.com/faq/) (see 'What would be a class name for an element inside another element?'), but it's fair to say tastes may differ, lmk 👍

jonathonherbert:
@paperboyo 

repeat: repeat sequence
emojis: 👍 😜 🆙 
moving around: this is rubbish, well spotted you and @akash1810 both – will correct 👍 
position: yes, a quick check of wikipedia tables and they fix it to the rhs of the column – would that make sense?
none: 'none' is just the usual workflow state – the order we get back from the server. It has to be expressed somehow, no? This may need a discussion
grouped by: the status, as before. Yes, could consider a global view also!

paperboyo:
> fix it to the rhs of the column

I only said left, because (I think!) it would make it clearer than right. Reasoning (weak!):
- there are no column dividers
- column width is fluid
Because of these two points, RHS would be less clear, I think (_does this triangle belong to value next to it, or this far-left thing?_).



> My thoughts are that this has to be expressed somehow, no?

Exactly. And stick a triangle to the column that represents that view. And get rid of _none_, then (clicking other columns **never** sets this sort). Like everywhere else (Gsheets etc).

> Yes, could consider a global view also!

Would be cool. In case we would allow grouping by different things, we would need some UI idea of how to express it…

public/lib/column-defaults.js
akash1810: `heading-sort-by` 👍 

This was a minor comment, please feel free to ignore too!

public/lib/column-defaults.js
jonathonherbert: (hmm, on further inspection would we ever want to sort by headings that aren't visible? IDK, but I'll add it as an option anyway)

paperboyo: > would we ever want to sort by headings that aren't visible

Interesting point! What happens if the active sort column is removed? I would say: this should be the only instance of thi _none_ (default sort state): sorting should revert (and be indicated against) default sort column. In case default sort column is itself removed… hmm… shiiiite. (next one right? haha)

jonathonherbert:
Re: left, only disadvantage – it moves column headings to the right of the row content below them, inconsistently (because some columns are sortable, and some aren't). The answer may be, make all the columns sortable! Which we may – but not yet 😁 

I'll try out both, bear with.

> And stick a triangle to the column that represents that view. 

That's the problem – which column is that? I don't know – and the server may change this without notice! We could find out and add something arbitrarily (It looks like `Working title / Headline` 😁), but really unless the client is in charge (we could make it in charge!) this is not something we can represent. Perhaps we should make the client in charge and sort by headline as a default.

The reason I held back on making this change is that other tables of this sort (I look at wikipedia again) do have this idea of a default, somehow 'unopinionated' sort, which users can change by clicking the heading.

paperboyo:
> only disadvantage – it moves column headings to the right

Shit, true. Possible to reserve space there? Or would it make some columns too wide and waste space>

> Perhaps we should make the client in charge and sort by headline as a default.

Is it seems the default anyway: this makes sense. 

> this idea of a default, somehow 'unopinionated' sort

From our own experience: we cannot figure out what it even is, this is confusing. Better if it’s indicated by a triangle, I think.

jonathonherbert:
Consensus on chat is that triangles next to column names is preferable at present 👍

PR #254, opened 2020-06-21
Fix section selection for modal so it works with custom desks

AWare:
## What does this change?

This changes the section dropdown in the create modal to work with custom desks. 
Previously it used the selected desk and filtered the sections off that, but the selection state is already *in* the section model. 

Section selection will be limited to selected sections.

## Images
Kapture 2020-06-21 at 14 05 57


paperboyo:
Feature request (going by the GIF in desc): if there is one and only on e Section to choose from the dropdown, would it make sense for it to be preselected?

AWare:
@paperboyo will do this next time I'm in workflow

paperboyo:
> @paperboyo will do this next time I'm in workflow

😘

PR #256, opened 2020-06-22
Add more sortable columns

jonathonherbert:
## What does this change?

Adds more sortable columns to workflow. Numbers now behave themselves – dates should too (and always have, I thought they'd need more work but they're all ISO8601 AFAICS).

sort-workflow-cols-2

This should include every column with heading text – we can add others, although it might be a bit of UX to figure out how to represent them.

Click through each column heading in workflow. Things should sort correctly on a per-section basis, including numbers and dates.

There's one caveat – some columns conditionally hide values if the content type isn't `article`. There are a few examples on CODE but I wasn't sure how common this was in PROD – if this turns out to be an issue we can revisit how these values are modelled.

### DIFF
--- a/public/components/content-list-item/content-list-item.js
+++ b/public/components/content-list-item/content-list-item.js
@@ -1,8 +1,8 @@
 
-var OPHAN_PATH = 'https://dashboard.ophan.co.uk/summary?path=/',
+const OPHAN_PATH = 'https://dashboard.ophan.co.uk/summary?path=/',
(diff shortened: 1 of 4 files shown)


paperboyo:
> a bit of UX to figure out how to represent them

A bit:
image

Yeah, bit shit, but consistent and better to:
- have them sortable as it’s potentially useful
- have them sortable as it’s more predictable and easier to internalise if you can sort all, instead of just some

Status makes no sense until we tackle grouping.

PR #257, opened 2020-06-23
Disable debug info to improve performances

mchv:
## What does this change?

The associated changes disable  debug data as by default AngularJS attaches information about binding and scopes to DOM nodes, and adds CSS classes to data-bound elements.

According to [angular documentation](https://code.angularjs.org/1.8.0/docs/guide/production):

>  you can disable this in production for a significant performance boost  

💥 ⚡ 

See similar changes:
 - [grid](https://github.com/guardian/grid/pull/1882)
 - [composer](https://github.com/guardian/flexible-content/pull/2545)

## How to review and test
- check that nothing break

From  @akash1810  `There is an argument that perf isn't down to angular, but rather our implementation`, so the test case in the discussion thread can be used to test if performance is better.

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -88,6 +88,8 @@ angular.module('workflow',
         $compileProvider.aHrefSanitizationWhitelist(
             sanitizeUrl(_wfConfig.indesignExportUrl)
         );
+        
+        $compileProvider.debugInfoEnabled(false);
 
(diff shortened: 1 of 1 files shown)


mchv:
@jonathonherbert or @akash1810 I let you merge and deploy as I will not be able to test this properly.

jonathonherbert:
Tested on CODE, I'll merge.

PR #258, opened 2020-06-24
Add the ability to sort by priority

jonathonherbert:
## What does this change?

Adds the ability to sort by priority.

sort-workflow-cols-3

To achieve this, two other changes were necessary –

- Moving the compactor icon from the priority column heading into the lhs sidebar, above the timezone selector
- Adding a secondary sort on the 'Working title' column when sorting by priority, to better reflect the order users expect things to be sorted by default.

The priority sort should work as expected – sorting first by priority, then by working title.
The compactor button, now moved, should work as it did.


public/components/compactor-toggle/_compactor-toggle.scss
```diff
@@ -0,0 +1,39 @@
...
+          0deg,
+          transparent,
+          transparent 1px,
+          $c-grey-400 1px,
+          $c-grey-400 2px
+  );
+
+  &--active {
+      background: repeating-linear-gradient(
+              0deg,
```
akash1810, line 32: sigh, we need a `.editorconfig` in this repo, 4 space indents is huge!

public/lib/column-defaults.js
```diff
@@ -64,13 +64,14 @@ var columnDefaults = [{
     // There's a good reason though, principally it's a wider column than the column to the left, notifier.
```
akash1810, line 64: I think this comment can go now?

PR #260, opened 2020-06-24
Provide sensible defaults on column sort

jonathonherbert:
## What does this change?

This applies the following defaults to the sortable workflow columns:

- Priority, first-click view ▲ (top priority on top, low bottom)
- Webtitle, first-click view ▲ (A at top, Z at bottom)
- Notes, first-click view ▲ (A at top, Z at bottom)
- Office, first-click view ▲ (A at top, Z at bottom)
- Deadline, first-click view ▼ (newest at top, oldest at bottom)
- Section, first-click view ▲ (A at top, Z at bottom)
- Webwords, first-click view ▲ (low nos. top, high nos. bottom) <--- for discussion?
- Printwords, first-click view ▲ (low nos. top, high nos. bottom) <--- for discussion?
- Publication location, first-click view ▲ (Booksection A–Z, then pubdate newest at top, then page no. low nos. top)
- Commission, first-click view ▲ (low nos. top, high nos. bottom)
- State, first-click view ▲ (State A–Z, then date newest top)
- Last modified, first-click view ▼ (newest at top, oldest at bottom)
- Last modified by, first-click view ▲ (A at top, Z at bottom)

There are two caveats –

- Sparsely populated fields like publication location, notes, state etc. treat no data as less-than some data, which leads to some emptier-than-expected first states. We can discuss how to deal with these – it might be better to treat empty fields according to the default sort order to push them downwards when we first click by default.
- With the current approach, 'Publication location' and 'State' might not quite reflect what's in the list (we can't break the data down into its parts and sort each part with different directions yet).

Click on the column headings – they should behave as above, notes above notwithstanding.

### DIFF
--- a/public/lib/column-defaults.js
+++ b/public/lib/column-defaults.js
@@ -34,22 +34,27 @@ import publicationLocationTemplate from "components/content-list-item/templates/
  *      colspan: number, // colspan of this field
  *      title: string, // title attribute contents for the column heading
  *      templateUrl: string // URL for the content-list-item template for this field
- *      isSortable: boolean // Can the column be sorted by clicking its header?
- *      sortField: string // The field to sort on, if different from `name`. Can be an item path, e.g. `a.nested.field`
+ *      isSortable: boolean = false // Can the column be sorted by clicking its header?
+ *      sortField?: string // The field to sort on, if different from `name`. Can be an item path, e.g. `a.nested.field`
+ *      defaultSortOrder?: 'asc' | 'desc' // The default sort order for the column on first click. Defaults to 'asc'
+ *      flipSortIconDirection: boolean = false // Flip the direction of the sort chevron relative to the sort order. Defaults to 'desc' -> ▼
  * }
  */
 
 var templateRoot = '/assets/components/content-list-item/templates/';
 
-const createSortTemplate = (sortField, labelHTML) => `
+const chevronUp = '&#9660;'
+const chevronDown = '&#9650;'
+
+const createSortTemplate = (sortField, labelHTML, flipSortIconDirection = false) => `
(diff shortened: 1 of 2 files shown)


public/lib/column-defaults.js
```diff
@@ -34,30 +34,39 @@ import publicationLocationTemplate from "components/content-list-item/templates/
...
+const createSortTemplate = (sortField, labelHTML, flipSortIconDirection = false) => `
     <div ng-click="toggleSortState('${sortField}')" class="content-list-head__heading-sort-by">
       ${labelHTML}
       <span
         class="content-list-head__heading-sort-indicator"
         ng-class="{invisible: !getSortDirection('${sortField}')}"
         ng-switch="getSortDirection('${sortField}')">
-        <span ng-switch-when="desc">&#9660;</span>
-        <span ng-switch-when="asc">&#9650;</span>
+        <span ng-switch-when="desc">${flipSortIconDirection ? chevronDown : chevronUp}</span>
```
akash1810, line 56: A ternary in a switch is a bit complicated. Can this be simplified somehow? Or add a comment to explain why it's needed?

jonathonherbert: Agreed – could farm this out to a selector in the controller, as with `getSortDirection()` above? 

Or better still, store these variables in the function body above and use the vars in the template?

Definitely a comment is warranted 👍

jonathonherbert: @akash1810 as this has hopefully been addressed, I'll merge this now, but lmk if additional changes would make things clearer 👍

akash1810: 👍 great work!

PR #261, opened 2020-07-01
Deadline datepicker

jennygrahamjones:
## What does this change?
Replaces the dropdown menu on the Deadline filter with a regular input element. 

The styling on the 'Choose date' label is not great, so I'd welcome any tips on how to do that (or any other styling) better. 

The picker itself will look [different depending on the browser you're viewing it in](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date).

Users can filter by any deadline date. The filter can be dismissed in the usual way, by clicking 'All'.

## Images
Screen Recording 2020-07-03 at 09 50 27

### DIFF
--- a/public/lib/filter-defaults.js
+++ b/public/lib/filter-defaults.js
@@ -131,6 +131,7 @@ var filterDefaults = function (statuses, wfFiltersService) {
                 const deadline = storedSelectedDate ? storedSelectedDate : '';
 
                 // Can our deadline be parsed into a date?
+                // If so, it's a custom date filter (rather than 'today', 'tomorrow' etc)
                 const parsedDate = moment(deadline)
                 if (parsedDate.isValid()) {
                     $scope.selectedFilters = ['customDate']
--- a/public/lib/filter-defaults.js
+++ b/public/lib/filter-defaults.js
@@ -1,3 +1,5 @@
+import moment from 'moment';
+
 /**
  * This returned array represents the default configuration of the sidebar filters list.
  * This is used by the components/sidebar-filter/sidebar-filter.js directive.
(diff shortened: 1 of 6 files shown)


AWare:
The weird arrow on choose date is from `.sidebar__filter-custom--customDate`
For the text I'd take a look at the class names nearby?

paperboyo:
This looks great! Some thoughts:
- could we prevent panel from resizing when dummy text is replaced by a date in the gif?
- as this is a native picker, do we know if monts<>days OS settings confusion will be handled correctly by us?
- [supah-minor] this chevron area looks like the tallest man on Earth: could we make it the height of the 1 row (next to date display)?
- [supah-minor] I would leave styling of `Choose date` as is. Or – maybe only bold (but not bigger and unindented).

jennygrahamjones:
> could we prevent panel from resizing when dummy text is replaced by a date in the gif?

I can't seem to replicate that happening now. Did you observe it in CODE? 

> as this is a native picker, do we know if months<>days OS settings confusion will be handled correctly by us?

From the docs, it looks like the value of a date input once picked is always parsed as yyyy-mm-dd to match [ISO 8601](https://www.iso.org/iso-8601-date-and-time-format.html). This should be browser/OS agnostic. 

> [supah-minor] this chevron area looks like the tallest man on Earth: could we make it the height of the 1 row (next to date display)?
> [supah-minor] I would leave styling of Choose date as is. Or – maybe only bold (but not bigger and unindented).

I've tried removing the chevon and resizing/styling the input. What do you think @paperboyo ?

Screen-Recording-2020-07-01-at-1

public/lib/filter-defaults.js
```diff
@@ -125,29 +127,26 @@ var filterDefaults = function (statuses, wfFiltersService) {
...
 
-                // ensure that the date from the URL is the same object as the
-                // one used in the Select drop-down, as its compared with ===
-                $scope.dateOptions.forEach(function (date) {
-                    if (date.isSame(selectedDate)) {
-                        selectedDate = date;
-                        $scope.selectedFilters = ['customDate'];
-                    }
-                });
+                // Can our deadline be parsed into a date?
```
jonathonherbert, line 133:
```suggestion
                // Can our deadline be parsed into a date?
                // If so, it's a custom date filter (rather than 'today', 'tomorrow' etc)
```

jonathonherbert: (Just to address why we care!)

paperboyo: I most probably misunderstood, but we *do want* to keep relative dates as they are immensely useful. Eg a bookmark can always show _tomorrow_ whatever day it’s clicked on. Actually, we may need more relative date choices (but here, I defer to John S.).

jennygrahamjones: Today, tomorrow, and weekend still behave as normal 👍

PR #264, opened 2020-07-03
Rename Publication location to Print location

paperboyo:
Ronseal. We are renaming this to be more inline with what Composer called it for years. Why column different from Management panel where it’s gonna be Print information? Cos in other places this may/will contain other print-related information in the future, while the column just holds info about location. And should, maybe, be three columns anyway…

- [x] tested on CODE


public/lib/column-defaults.js
```diff
@@ -258,8 +258,8 @@ const columnDefaults = [{
     sortField: 'printWordCount'
 },{
     name: 'publicationlocation',
```
akash1810, line 260: Could we change the `name` field too? And also rename the html template above? Having them called `publicationLocation` but displaying them as `Print location` will very soon become confusing.

Happy to pair on this.

paperboyo: Done.

PR #268, opened 2020-07-13
Sort by multiple fields

SHession:
## What does this change?

This PR provides the ability to sort workflow fields by multiple fields specified in the column-default.js files. A dev can specify the order of fields to sort by and a direction for each of those fields. 

This additional functionality allows us to sort by the more complicated sort specified in the [trello card](https://trello.com/c/HjpEVCmx/270-correct-sorting-for-multi-value-fields). See the print location column in the column-default.js to see how it has been setup.


philmcmahon:
Today I learned workflow has sort functoinality! Amazing!

SHession:
> Today I learned workflow has sort functoinality! Amazing!

We're truly living in a golden age!

PR #280, opened 2020-08-11
Add print modified-by and status fields

aug24:
## What does this change?
Adds two new elements to the workflow dash: "last modified in print by", and "status in print".

## How to test

Deploy this branch into CODE, along with https://github.com/guardian/workflow/pull/1070. Push [example Octopus-shaped data](https://gist.github.com/jennygrahamjones/3df30de9d947eefc86f767c6f4566104) (for Composer content tracked in Workflow) onto the octopus-integration-CODE Kinesis stream. Observe that Workflow frontend displays the relevant values.

Fields are present, correctly populated, and look nice.

N/A

## Images
N/A

### DIFF
--- a/common-lib/src/main/scala/models/Stub.scala
+++ b/common-lib/src/main/scala/models/Stub.scala
@@ -11,7 +11,6 @@ import io.circe.parser.decode
 import io.circe.syntax._
 import org.joda.time.{DateTime, LocalDate}
 import DateFormat._ // Required for serialisation / deserialisation of DateTime
-
 import scala.collection.immutable
 
 case class ExternalData(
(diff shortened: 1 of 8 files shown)


common-lib/src/main/scala/models/Stub.scala
```diff
@@ -46,7 +46,9 @@ case class ExternalData(
                          actualNewspaperPublicationDate: Option[LocalDate] = None,
                          // Description enriched for use by WF front end client code.
                          shortActualPrintLocationDescription: Option[String] = None,
-                         longActualPrintLocationDescription: Option[String] = None) {
+                         longActualPrintLocationDescription: Option[String] = None,
+                         statusInPrint: Option[String] = None,
```
jennygrahamjones, line 50:
```suggestion
                         statusInPrint: Status = Status.Writers,
```

I'm suggesting in https://github.com/guardian/workflow/pull/1070/files#diff-f5d71e0998bbbbf694aebeb923220c87 that we make this match the existing `status` field in terms of it's type.

PR #284, opened 2020-08-13
Add integration test app to allowed origins

stephengeller:
## What does this change?

This adds the ability to add allowed origins (namely the integration test origins) as a Cors allowed origin, mirroring the implementation in the Grid.

This change allows us to programmatically clean up after ourselves in the integration tests by deleting test content via the API.

We can cleanly delete content via the API from within our integration tests (PR to be merged once this one is).


stephengeller:
Changes to the config files have already been made, and tested on CODE successfully.

PR #286, opened 2020-08-17
Add needs picture desk

aug24:
## What does this change?
Adds a flag column for picture desk status.  Needs/Approved, exactly as per Legal.
Similarly, adds edit capability for that flag in the management pane.

## How to test
CODE

Can set and view icons 

Yes.


.java-version
```diff
@@ -1 +1 @@
-1.8.0.232
+1.8.0.242
```
jennygrahamjones, line 1: Was there a particular issue that prompted this?

public/components/content-list-item/templates/needsPictureDesk.html
```diff
@@ -0,0 +1,6 @@
+<td class="content-list-item__field--needs-picture-desk" title="Needs Picture Desk: {{ contentItem.needsPictureDesk.toLowerCase() }}">
+
+    <i ng-if="contentItem.needsPictureDesk == 'REQUIRED' || contentItem.needsPictureDesk == 'COMPLETE'" class="content-list-item__icon--incopy content-list-item__icon--inline" wf-icon="needs-picture-desk"></i>
```
jennygrahamjones, line 3: Not strictly related to this PR, but is it worth switching out `content-list-item__icon--incopy` for something like `content-list-item__icon--needs...`? At the minute, both this and `needsLegal` are using styling that seems to be tied to another icon.

.java-version
aug24: No, it's a mistake.  Reverted.

public/components/content-list-item/templates/needsPictureDesk.html
aug24: Done

jennygrahamjones (approved):

PR #287, opened 2020-08-19
Allow users to change the Section an atom is assigned to

jennygrahamjones:
## What does this change?
We've had feedback that it would be useful to be able to adjust the 'Section' a Chart atom is assigned to after the atom is created. Currently, there's no element in the UI of either Workflow or Atom Workshop that will allow you to change this value. This can be problematic for users if they accidentally select the wrong Section during creation or just want to change it at a later date.

## How to test
Run this branch locally, with the ssh tunnel to CODE setup, or deploy the branch to CODE. Click on a Chart atom and observe that the 'Section' dropdown is present in the Furniture tab of the drawer. Use the dropdown to change the value. Open Atom Workshop in CODE and confirm that the value has also changed in the Workflow info panel there.

It's possible to change the 'Section' an atom is assigned to.

This affects all atoms that aren't media atoms. I'm not sure if that's a problem?

## Images
Screenshot 2020-08-19 at 16 42 09

### DIFF
--- a/public/components/content-list-drawer/content-list-drawer.html
+++ b/public/components/content-list-drawer/content-list-drawer.html
@@ -318,6 +318,10 @@
                                 </span>
                             </li>
                             <li class="drawer__item">
+                                <p class="drawer__item-title">Section</p>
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
Brilliant. Looks good when tested per spec! ✨

PR #288, opened 2020-08-20
Changes to Create Article modal

jennygrahamjones:
## What does this change?
1. The collapsible 'Advanced' section of the Create Article modal is now removed, with all elements of the modal visible.
2. The 'Picture Desk' menu is visible under the 'Legal' menu.

## How to test
Run the branch locally with the SSH tunnel to CODE set up. Create a new Article and observe the modal appears as in the 'After' image below, with all options available. Pick some options from the Legal and Picture Desk menus. Create your Article. Observe the correct info is displayed on the dashboard.

More usage of the options that were previously hidden in the Advanced section. 

n/a

## Images
Before:
Screenshot 2020-08-20 at 12 00 04

After (screenshot taken on local rather than CODE, so some menus are not populated/local errors displayed):
Screenshot 2020-08-20 at 11 52 47

### DIFF
--- a/public/components/stub-modal/stub-modal.html
+++ b/public/components/stub-modal/stub-modal.html
@@ -66,16 +66,6 @@
                     </div>
                 </div>
             </div>
-        </div>
-
-    </div>
-    <div class="modal-body modal-body__header" ng-click="expanded = !expanded">
-        <a href="" class="modal-body__header-expander" ng-class="{'modal-body modal-body__header-expander--open': expanded}">
-            Advanced
-        </a>
-    </div>
-    <div class="modal-body modal-body__advanced" ng-class="{'modal-body__advanced--open' : expanded}">
-        <div class="form-group">
             <div wf-date-time-picker label="Production deadline" help-text="true" ng-model="stub.due"></div>
             <div class="form-horizontal clearfix">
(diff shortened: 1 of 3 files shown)


public/components/stub-modal/stub-modal.html
```diff
@@ -67,14 +67,6 @@ <h4 class="modal-title">{{ modalTitle }}</h4>
                 </div>
             </div>
         </div>
-
-    </div>
-    <div class="modal-body modal-body__header" ng-click="expanded = !expanded">
```
jonathonherbert, line 72: Just to double check, is this 'expanded' variable set in the scope anywhere? Does it need to be removed?

(It's possible to set this in the template only, so it may be this is the only reference)

jennygrahamjones: As far as I can see, this is the only place it's set or referenced. There's no mention of it in the corresponding JS file.

jonathonherbert (approved):
Presume that this has had the relevant 👀  eyes from editorial people, code looks good, tested locally, works as expected 👍

jennygrahamjones:
I've added the do-not-merge label until we get confirmation from CP to go ahead with these changes.

PR #293, opened 2020-08-24
hasPrintInfo to mean page no

paperboyo:
Depends on (and explained in): https://github.com/guardian/workflow/pull/1075

image


jennygrahamjones:
Will this mean changing the `Print location` column name too?

paperboyo:
Very good question! I think not, as the filter acts on the subset of Print location (only book info, not page no. and date)?

aug24 (approved):

paperboyo:
This is to change, explanation [here](https://github.com/guardian/workflow/pull/1075#issuecomment-680797627).

jonathonherbert (approved):

PR #295, opened 2020-08-25
Only call for content once at startup

stephengeller:
## What does this change?

Thanks to [this PR](https://github.com/guardian/workflow-frontend/pull/281), a request for content won't render if another request has occured since the initial request went out. The issue is that, when starting up Workflow, 3 requests for `content` get made. These are:

1. An intentional call for content from the `startPolling` function
2. The instantiation of the `selectedDesk` listener, which invokes the code within the callback
3. The instantiation of the `selectedView` listener, which does the same as above

Since each call to `content` returns large payloads (around 11mb on PROD), this results in Workflow waiting for the third call to respond before rendering anything.

 This PR ensures that the watchers only get invoked if there is a difference between the old and new values, which there isn't when the watcher gets invoked. This should offer significant performance improvements to Workflow at startup. 

## How to test

Spin up Workflow / go to CODE, open the Network tab and look for calls to `content` at startup. You should see there's only one, rather than the three before.

Loading articles in Workflow at startup takes 1/3rd of the time that it currently takes.

## Images

BEFORE:
image

AFTER:
image

### DIFF
--- a/public/lib/filters-service.js
+++ b/public/lib/filters-service.js
@@ -31,7 +31,6 @@ angular.module('wfFiltersService', ['wfDateService', 'wfTrustedHtml'])
 
                 $rootScope.$on('filtersChanged.section', function(event, data) {
                     self.update('section', data);
-                    console.log('filtersChanged.section')
                     $rootScope.$broadcast('getContent');
                 });
 
@@ -143,7 +142,6 @@ angular.module('wfFiltersService', ['wfDateService', 'wfTrustedHtml'])
 
                 $rootScope.$on('filtersChanged.view', function(event, data) {
                     self.update('view', data);
-                    console.log('filtersChanged.view')
                     $rootScope.$broadcast('getContent');
                 });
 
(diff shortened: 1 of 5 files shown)


public/lib/filters-service.js
```diff
@@ -49,7 +49,7 @@ angular.module('wfFiltersService', ['wfDateService', 'wfTrustedHtml'])
                     $rootScope.$broadcast('getContent');
                 });
 
-                $rootScope.$on('filtersChanged.deadline', function (event, data) { 
+                $rootScope.$on('filtersChanged.deadline', function (event, data) {
```
jennygrahamjones, line 52: Are there any expected changes here, or is this a product of linting?

jennygrahamjones (approved):
Thanks for the thorough PR description! Looks great on CODE 🚀

public/lib/filters-service.js
stephengeller: Just an accidental extra space from linting, no expected changes!

PR #297, opened 2020-08-26
Rename Last modified by to Last user…

paperboyo:
…mostly for brevity (columns are soooo wide). I wanted to be a good boy and not do it only in the UI, but it’s entirely possible it was a stupid move.


paperboyo:
Will just do UI instead… closing.

PR #298, opened 2020-08-26
Add data-cy selector for content list tables

stephengeller:
## What does this change?

This adds another selector for content list tables, so that Cypress can assert that an article has moved from one table to another after changing the status

### DIFF
--- a/public/components/content-list/content-list.html
+++ b/public/components/content-list/content-list.html
@@ -17,7 +17,7 @@
         </tr>
     </thead>
 
-    <tbody class="content-list__item-container" ng-repeat="group in content track by group.name" content-list-item-container>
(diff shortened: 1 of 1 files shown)


public/components/content-list/content-list.html
```diff
@@ -17,7 +17,7 @@
         </tr>
     </thead>
 
-    <tbody class="content-list__item-container" ng-repeat="group in content track by group.name" content-list-item-container>
+    <tbody data-cy="content-list-{{group.name}}" class="content-list__item-container" ng-repeat="group in content track by group.name" content-list-item-container>
```
stephengeller, line 20: Resolves to this: image

jennygrahamjones (approved):

PR #299, opened 2020-08-26
Rename Last modified by to Last user, take 2

paperboyo:
For brevity only. Replaces https://github.com/guardian/workflow-frontend/pull/297.


jennygrahamjones:
Just to double-check, the intention for this is just that it will rename the column names?

paperboyo:
> intention for this

Yep, just column headings and their counterparts in Column picker. Management panel deals with that differently.

jennygrahamjones (approved):

PR #301, opened 2020-09-02
Adjust presence opacity depending on user location

jennygrahamjones:
This PR introduces a small change to the way that Presence is handled in Workflow. We can use the `location` value provided by the Presence service to determine whether or not the user is actively editing the body copy. If the `location` is 'body', the Presence indicator will display as normal. Otherwise, the indicators opacity will be reduced.

## How to test
Run Workflow, Composer (using https://github.com/guardian/flexible-content/pull/3595), and the Presence services locally. Open a piece of Composer content. Observe the Presence indicator in Workflow and that it changes opacity depending on whether you are actively editing the body copy, or have clicked the 'Save and close' button and are no longer editing the body.

Alternatively, deploy this branch of Workflow to CODE and repeat steps above.

Users of Workflow have a better idea of when their colleagues are actively working on some content and when they have just left a browser tab open.

n/a?

## Images
Screen Recording 2020-09-02 at 15 51 15

Tooltips:

Screen Recording 2020-09-04 at 15 39 47

### DIFF
--- a/public/components/content-list-drawer/content-list-drawer.html
+++ b/public/components/content-list-drawer/content-list-drawer.html
@@ -503,7 +503,7 @@
                         </li>
                         <li class="drawer__item">
                             <p class="drawer__item-title">In use by</p>
-                            <wf-presence-indicators presence-id="contentItem.composerId" in-drawer="true" ng-if="contentItem.composerId"></wf-presence-indicators>
+                            <wf-presence-indicators presence-id="contentItem.composerId" dont-display-idle="contentItem.contentType == 'liveblog'" in-drawer="true" ng-if="contentItem.composerId"></wf-presence-indicators>
                             <wf-presence-indicators presence-id="'media-' + contentItem.item.editorId" in-drawer="true" ng-if="contentItem.contentType === 'media'"></wf-presence-indicators>
                         </li>
                         <li class="drawer__item">
--- a/public/components/presence-indicator/presence-indicators.html
+++ b/public/components/presence-indicator/presence-indicators.html
@@ -6,6 +6,6 @@
         ng-if="presence.status === 'present' || presence.status === 'idle'"
         ng-class="'content-list-item__presence--' + presence.status">
         <a class="content-list-item__icon--presence"
-           href="mailto:{{ presence.email }}" title="{{ inDrawer ? presence.email : presence.longText }}">{{ inDrawer ? presence.longText : presence.indicatorText }}</a>
+           href="mailto:{{ presence.email }}" title="{{ inDrawer ? presence.shorTtitle : presence.longTitle }}">{{ inDrawer ? presence.longText : presence.indicatorText }}</a>
     </li>
 </ul>
--- a/public/components/presence-indicator/presence-indicators.html
+++ b/public/components/presence-indicator/presence-indicators.html
@@ -3,7 +3,7 @@
     <span ng-show="(!presences.length || presences.length === 1 && presences[0].status === 'free') && inDrawer" class="drawer__section-data-row drawer__section-data-row--coming-soon">Nobody</span>
 
     <li ng-repeat="presence in presences track by presence.email"
-        ng-if="presence.status === 'present'"
+        ng-if="presence.status === 'present' || presence.status === 'idle'"
         ng-class="'content-list-item__presence--' + presence.status">
         <a class="content-list-item__icon--presence"
            href="mailto:{{ presence.email }}" title="{{ inDrawer ? presence.email : presence.longText }}">{{ inDrawer ? presence.longText : presence.indicatorText }}</a>
--- a/public/components/presence-indicator/presence-indicators.html
+++ b/public/components/presence-indicator/presence-indicators.html
@@ -6,6 +6,6 @@
         ng-if="presence.status === 'present' || presence.status === 'idle'"
         ng-class="'content-list-item__presence--' + presence.status">
         <a class="content-list-item__icon--presence"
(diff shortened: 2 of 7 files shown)


public/components/presence-indicator/presence-indicators.js
```diff
@@ -20,12 +20,22 @@ function wfPresenceIndicatorsDirective ($rootScope, wfPresenceService,
...
+                            const presenceObject = {
+                                indicatorText:
+                                    (person.firstName.charAt(0) + person.lastName.charAt(0)).toUpperCase(),
                                 longText: [person.firstName, person.lastName].join(" "),
-                                email: person.email,
-                                status: "present" };
+                                email: person.email
+                            };
+
+                            if(currentState[0].location === "body") {
```
jennygrahamjones, line 32: I'm not sure if there are other values that `location` can have. When you click 'Save and close' in Composer, leaving the `body` location, the Presence client still registers you as being present in the Article but with an empty location value.

public/components/presence-indicator/presence-indicators.js
```diff
@@ -20,12 +20,22 @@ function wfPresenceIndicatorsDirective ($rootScope, wfPresenceService,
...
+                                // the user is actively editing the body
+                                return {...presenceObject, ...{status: "present"}};
+                            } else {
+                                // the user is not editing the body, has clicked 'Save and close'
+                                return {...presenceObject, ...{status: "idle"}};
```
jennygrahamjones, line 37: Very much open to suggestions on a better name than "idle"

jennygrahamjones: Perhaps that's a good fit, yes - although notably, the Presence client doesn't appear to register any difference between editing and not editing furniture.

paperboyo: > client doesn't appear to register any difference

This may be something we would want to address together with this bigger change that will allow multiple presence indicators in Liveblog blocks…

For now, this is great (as currently, editing furniture can happen concurrently to body editing).

SHession (commented):
I tested this on CODE and it works brilliantly for articles but I word argue not as well for other types of content. Other content types managed in workflow (media atoms, charts, etc.), only show the greyed out presence indicator. I would suggest these should behave how they previously did, with the new functionality for articles only.

paperboyo:
> new functionality for articles only

Great points about atoms and charts. But should it stay like this for anything edited in Composer (liveblogs, galleries etc) too?

> adding a tooltip

💯

SHession:
> Great points about atoms and charts. But should it stay like this for anything edited in Composer (liveblogs, galleries etc) too?

Testing this on CODE, it works okay for galleries but not for liveblogs. This is because your presence icon doesn't track what content you're editing in a liveblog. This was attempted but found to be tricky due to presence not currently being able to handle a user looking at multiple items within one tab. Meaning, the icons for liveblogs always appears semi-transparent.

jennygrahamjones:
Thanks @SHession! I'll revisit how I've implemented this.

jennygrahamjones:
I've made some small changes and in conjunction with https://github.com/guardian/flexible-content/pull/3595 I think the indicator should now be behaving correctly, with regards to opacity, for the other content you've mentioned, @SHession

SHession (dismissed):
Tested this on CODE and it looks good. The opacity is always reduced for liveblogs, I'm not sure if that is an issue?

A tooltip or similar would still be nice, but I don't think it's a blocker.

jennygrahamjones:
I'll see if I can add something to https://github.com/guardian/flexible-content/pull/3595 for liveblogs, and look at that tooltip too.

public/components/presence-indicator/presence-indicators.html
```diff
@@ -6,6 +6,6 @@
         ng-if="presence.status === 'present' || presence.status === 'idle'"
         ng-class="'content-list-item__presence--' + presence.status">
         <a class="content-list-item__icon--presence"
-           href="mailto:{{ presence.email }}" title="{{ inDrawer ? presence.email : presence.longText }}">{{ inDrawer ? presence.longText : presence.indicatorText }}</a>
+           href="mailto:{{ presence.email }}" title="{{ inDrawer ? presence.shorTtitle : presence.longTitle }}">{{ inDrawer ? presence.longText : presence.indicatorText }}</a>
```
SHession, line 9: This looks like a typo:

```suggestion
           href="mailto:{{ presence.email }}" title="{{ inDrawer ? presence.shortTitle : presence.longTitle }}">{{ inDrawer ? presence.longText : presence.indicatorText }}</a>
```

public/components/content-list-drawer/content-list-drawer.html
```diff
@@ -503,7 +503,7 @@
                         </li>
                         <li class="drawer__item">
                             <p class="drawer__item-title">In use by</p>
-                            <wf-presence-indicators presence-id="contentItem.composerId" in-drawer="true" ng-if="contentItem.composerId"></wf-presence-indicators>
+                            <wf-presence-indicators presence-id="contentItem.composerId" dont-display-idle="contentItem.contentType == 'liveblog'" in-drawer="true" ng-if="contentItem.composerId"></wf-presence-indicators>
```
jennygrahamjones, line 506: Would be keen to know if anyone has a better idea of what to name this new value.

It's being used to flag that live blogs should be treated differently to other content, in that it should always display the 'editing in body' Presence indicator.

jonathonherbert: Hm, it's clear to me, although happy to hear from others 👍

jonathonherbert (approved):
Tested on CODE, behaves as expected 👍

jennygrahamjones:
[Diferrentiate body-editing and just-looking Presence indicator in Workflow](https://trello.com/c/kc015R3T/289-diferrentiate-body-editing-and-just-looking-presence-indicator-in-workflow)

PR #302, opened 2020-09-08
Display all of my initials please :)

jennygrahamjones:
## What does this change?
I've been looking at Presence indicators _a lot_ recently, so noticed this small descrepancy between the way we display an indicator of who content is assigned to and who is currently viewing the content. If I assign some content to myself, it is assigned to `JGJ` but if I view it, it's being viewed by `JG`. It looks strange, seeing the two side-by-side. At a glance, someone might think JG and JGJ are two different people.

Caveat: I don't know if this is the best way to do this with Angular.

## How to test
Have a name with more than two parts to it and use Workflow :)

Presence indicator behaves the same as asignee indicator in terms of the letters it displays.

n/a?

## Images
Before
Screenshot 2020-09-08 at 14 44 32
After
Screenshot 2020-09-08 at 14 44 58


public/components/content-list-item/content-list-item.js
```diff
@@ -2,6 +2,16 @@
 const OPHAN_PATH = 'https://dashboard.ophan.co.uk/summary?path=/',
     LIVE_PATH = 'http://www.theguardian.com/';
 
+export function toInitials(str) {
```
SHession, line 5: There may be a better place for this function live, possibly a helper library somewhere or a separate file.

jennygrahamjones: Yeah, I did think about that, since it's now used in two places. I'll have a look and see if there is an existing helpers or utils file that it would make sense in.

akash1810:
Would we want to show all initials across the suite of Tools to avoid confusion?

jennygrahamjones:
> Would we want to show all initials across the suite of Tools to avoid confusion?

Yes, it would make sense to be consistent wherever we use a persons initials, I think. This change would make Workflow consistent with itself, but admitedly not with Composer and whichever other tools implement Presence.

paperboyo:
> across the suite of Tools

If it’s only Assign in Workflow that uses middle names (?), then maybe making it not use it, instead of making all the other places use it will be a cheaper way to provide consistency (plus, ppl with 3 letters, but used to 2 letters will be less confused). But I don’t know how it is in other places: my middle name is known only to me bank.

jennygrahamjones:
@paperboyo This wouldn't relate to middle names, but people who have more than one part to their first name or surname e.g. a double-barrelled name like myself and @srbd, or someone with a [compound surname](https://en.wikipedia.org/wiki/Surname#Compound_surnames). 

Happy to switch Workflow to display only two initials in all instances though, if that's preferable! I just want to be consistent 😅

paperboyo:
For the record, we’ve checked MAM, AW and Composer with Jenny and all show just two letters… so… :-)

SHession (approved):

PR #316, opened 2020-11-02
Don't use the CAPI key as Panda system

sihil:
## What does this change?
An error during a Play upgrade has meant that workflow's panda ID is the CAPI key that workflow uses. This is, umm, a mistake. This should be a static meaningful ID. Fortunately it is static so it is still working, but this makes it meaningful again.

## How to test
This should be a no-op, although some users might need to refresh panda.

`workflow` is the panda ID rather than a seemingly random UUID.

Some users might need to refresh panda but when testing in CODE I saw no impact.


akash1810 (approved):
Thanks!

PR #317, opened 2020-11-27
Hide the 'View versions' button for content that isn't in Composer

jennygrahamjones:
## What does this change?
A small change based on observing a user interact with Workflow this morning. The user wanted to see who had worked on a piece of Video content, so clicked the 'View versions' button in the management panel. This button passes the Composer ID on to Restorer, to display a version/snapshot history. If there is no Composer ID, the value `null` is passed in it's place. This PR hides the 'View versions' button if `null` is present in the Restorer URL.

## How to test
In Workflow, find some content of the following types (Content Type of Video or Atom Type of Media).

Screenshot 2020-11-27 at 09 26 58

Check whether the content has an associated Composer page. If it doesn't, there will be a 'Create Video Page' button visible in Media Atom Maker.

Screenshot 2020-11-27 at 09 41 23

Running this branch locally or in CODE, observe that the 'View versions' button should not be visible in the management panel for any content that does not have a Composer page.

1. Will this be confusing to users who are used to seeing this button here? Arguably it is better UX, because they won't click it only to discover that it doesn't work.

1. The conditional check for `null` in the Restorer URL is quite basic. Is there a better attribute, on the `contentItem` model perhaps, that could give us an indication about whether the content will be found in Restorer? I'd originally thought to use the content type, but given a video may start off without an ID and then later acquire one, we can't rely on the content type.

### DIFF
--- a/public/components/content-list-drawer/content-list-drawer.html
+++ b/public/components/content-list-drawer/content-list-drawer.html
@@ -619,7 +619,8 @@
             <div class="drawer__toolbar-item drawer__toolbar-item--spacer">
                 &nbsp;
             </div>
-            <span class="drawer__toolbar-item drawer__toolbar-item--action">
+            <!-- if the content doesn't have a Composer page, it won't be accessible in Restorer -->
+            <span ng:if="!composerRestorerUrl.includes('null')" class="drawer__toolbar-item drawer__toolbar-item--action">
                 <a href="{{ composerRestorerUrl }}" target="_blank" class="drawer__item-content" title="View versions">
                     <i class="drawer__icon drawer__toolbar-icon" wf-icon="view-versions"></i>
                     View versions


paperboyo:
Alternative would be to take user to the management tab in MAM:

image

But this is unworkable, coz MAM doesn’t allow deep-linking to the pane (something always worthy to remember about when designing new tools).

Related would be for Workflow to use MAM’s (and other non-Composer editors’) value for Last user (currently empty).

Regardless, I think this is still useful as is, because showing a button leading to nowhere is daft.

jennygrahamjones:
> Related would be for Workflow to use MAM’s (and other non-Composer editors’) value for Last user (currently empty).

👍  That would definitely be an improvement. It's interesting though that on this occasion, I don't think I saw the user relying on the 'last modified...' columns or info for help. They seemed to be relying on Restorer as their way to answer "who has touched this piece of content?" - hopefully other projects we are working on will give visibility of that history for tools that don't have the same versioning in place 🤞

mbarton:
Interestingly Media Atom Maker used to have an audit trail in each atom but [we removed it](https://github.com/guardian/media-atom-maker/pull/868#issuecomment-480065134). I agree a centralised trail would be 💯

SHession (approved):
LGTM

PR #320, opened 2021-01-04
Add pinboard column to dashboard

jennygrahamjones:
## What does this change?
Provides example integration with pinboard by adding a new column that shows the number of pinboard messages, and can be clicked to open the pinboard itself. Some food for thought for how it might look/work.

## Images
pinboard-workflow(1)

### DIFF
--- a/app/views/app.scala.html
+++ b/app/views/app.scala.html
@@ -25,4 +25,5 @@
     </script>
   }
   <script async src="https://apis.google.com/js/client.js?onload=init"></script>
+    <script async src="https://pinboard.local.dev-gutools.co.uk/pinboard.loader.js"></script>
 }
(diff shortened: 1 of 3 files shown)


app/views/app.scala.html
```diff
@@ -25,4 +25,5 @@
     </script>
   }
   <script async src="https://apis.google.com/js/client.js?onload=init"></script>
+    <script async src="https://pinboard.local.dev-gutools.co.uk/pinboard.loader.js"></script>
```
jennygrahamjones, line 28: Need to pull this value out of config.

public/lib/column-defaults.js
```diff
@@ -122,6 +123,16 @@ const columnDefaults = [{
...
+},{
+    name: 'pinboard',
+    prettyName: 'Pinboard',
+    labelHTML: 'Pinboard',
+    colspan: 1,
+    title: '',
+    templateUrl: templateRoot + 'pinboard.html',
+    template: pinboardTemplate,
+    active: true,
+    isSortable: true,
```
jennygrahamjones, line 135: Not sure whether we would want this to be sortable.

twrichards:
done for real in https://github.com/guardian/workflow-frontend/pull/385 and https://github.com/guardian/pinboard/pull/219

PR #324, opened 2021-02-05
Adds support for the new presence locations field

SHession:
## What does this change?

This PR adds support for the new [Presence locations](https://github.com/guardian/presence-indicator/pull/114) in the icons which appears against stubs.

This change should facilitate backwards compatibility with the existing functionality whilst supporting the new field.

it is also a slight change as editing the furniture will now count as actively editing a document. Previously, this only applied to editing the document.

## How to test

This change would ideally be tested against the Presence updates with no `locations` and then with some `locations` to ensure it works before and after the changes. Currently no `locations` will be stored in Presence until Presence clients start adding them i.e. in [this PR](https://github.com/guardian/flexible-content/pull/3764).

The Presence behaviour continues to work as before and also understands users editing multiple locations.

### DIFF
--- a/public/components/presence-indicator/presence-indicators.js
+++ b/public/components/presence-indicator/presence-indicators.js
@@ -30,16 +30,18 @@ function wfPresenceIndicatorsDirective($rootScope, wfPresenceService,
                                 email: person.email
                             };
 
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
Tested in CODE, new indicators respond to entry on both body and furniture 👍

PR #325, opened 2021-02-10
Adjust presence display logic to hide idle and show furniture as semi-transparent

SHession:
## What does this change?

This PR use the multiple location Presence work to display the icons differently depending on the users location.

The behaviour is now:

1. In body -> Opaque icon
2. In furniture -> Semi-transparent icon
3. In neither body or furniture -> no icon

Note: This behaviour is also featured in the content drawer, in the management pane. 

Note 2: Liveblogs are currently not supported but could be added. This would require a change to Composer to more clearly identify liveblog blocks in the locations field.

## How to test

This is probably best tested on CODE in conjunction with https://github.com/guardian/flexible-content/pull/3764. However it is also possible to test locally but requires Presence, Workflow, Workflow-Frontend & Composer to all be running.

Workflow users are able to clearly identify who is working in a given piece of content, and what they're doing.

## Images

WORKFLOW_PRESENCE

### DIFF
--- a/public/components/presence-indicator/presence-indicators.js
+++ b/public/components/presence-indicator/presence-indicators.js
@@ -36,32 +36,29 @@ function wfPresenceIndicatorsDirective($rootScope, wfPresenceService,
 
                             if (editingBodyLocations.some((l) => currentLocations.includes(l)) || $scope.dontDisplayIdle) {
                                 return {
-                                    ...presenceObject, ...{
-                                        status: "present",
-                                        longTitle: [presenceObject.longText, "editing body"].join(" - "),
-                                        shortTitle: [presenceObject.email, "editing body"].join(" - "),
-                                        iconPrecedence: 1
-                                    }
+                                    ...presenceObject, 
+                                    status: "present",
+                                    longTitle: [presenceObject.longText, "editing body"].join(" - "),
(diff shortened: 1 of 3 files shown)


public/components/presence-indicator/presence-indicators.js
```diff
@@ -29,30 +29,37 @@ function wfPresenceIndicatorsDirective($rootScope, wfPresenceService,
...
-                                        shortTitle: [presenceObject.email, "editing"].join(" - "),
+                                        longTitle: [presenceObject.longText, "editing body"].join(" - "),
+                                        shortTitle: [presenceObject.email, "editing body"].join(" - "),
                                         iconPrecedence: 1
                                     }
                                 };
+                            } else if(editingFurnitureLocations.some((l) => currentLocations.includes(l))) {
+                                // the user is not editing the body, has clicked 'Save and close'
+                                return {
+                                    ...presenceObject, ...{
```
jonathonherbert, line 48: IIUC we should be able to flatten this spread into the parent object.

SHession: Well spotted, I was following the convention in that file and missed that.

jonathonherbert (approved):

PR #328, opened 2021-02-15
Checks data value before accessing

SHession:
## What does this change?

This PR attempts to fix a long standing issue in [sentry with millions of logs](https://sentry.io/organizations/the-guardian/issues/1860318059/?project=33134&query=is%3Aunresolved). It seems to be a simply case of accessing a value on an undefined object. Fixing this issue should hopefully remove some of the noise from Sentry.

## How to test

The changes should be a NO-OP and Workflow should behave as normal. 

We can measure success by watching the sentry logs for this issue drop off.

### DIFF
--- a/public/lib/preferences-service.js
+++ b/public/lib/preferences-service.js
@@ -127,7 +127,7 @@ angular.module('wfPreferencesService', [])
                         }
                     } else {
                         return this.prefsPromise.then(function resolve (data) {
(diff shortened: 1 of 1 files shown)


tjsilver (approved):
Code makes sense (although I was unable to see the Sentry logs - have requested access for the future).

PR #347, opened 2021-12-02
updates vulnerable dependancies via snyk

dskamiotis:
## What does this change?

This addresses vulnerabilities presented in Snyk:
image

## Libraries updated
Some key libraries were updated:
- angular
- lodash
- ua-parser-js
- jquery-ui

## How to test

Deploy to CODE and check https://workflow.code.dev-gutools.co.uk/dashboard to ensure functioning as expected.
This also be done as instructed [here](https://workflow.code.dev-gutools.co.uk/dashboard)

Run tests within `sbt test` 

Deploy branch to CODE tested in `https://workflow.code.dev-gutools.co.uk/dashboard`
- Created new article successfully
- Modified writers status
image

## How can we measure success?

We can see the number and level of vulnerabilies in [Snyk](https://app.snyk.io/org/guardian/project/581b49a4-bdbb-4d3f-899f-b7521c73ce24)

Check vulnerabilities after updates using Snyk CLI command `snyk test --all-projects`:
Before updates:
image
After updates:
image

That workflow-frontend is working as expected https://workflow.code.dev-gutools.co.uk/dashboard
## Have we considered potential risks?

Risks pertaining to the updates causing a breaking change are mitigated through thorough testing, see above.

PR #349, opened 2021-12-13
Revert "updates vulnerable dependancies via snyk"

SHession:
Reverts guardian/workflow-frontend#347. This is due to user reported issue with editing notes. We have also seen issue in Sentry and our logs.

Reverting until we address these concerns.


dskamiotis (approved):
Revert well justified

PR #352, opened 2021-12-16
Upgrade sbt to 1.5.7 and Play to 2.8.11

rhystmills:
## What does this change?

This PR upgrades the Workflow SBT and Scala Play versions to the latest. This is necessary to mitigate [security issues identified in log4j2](https://github.com/sbt/sbt/releases/tag/v1.5.6).

This PR makes several changes:

- Upgrades SBT to v1.5.7
- Upgrades Scala Play to v2.8.11
- Upgrades AWS libraries to v1.12.129
- Performs necessary migration steps for these upgrade to work

[Editorial tools working document](https://docs.google.com/document/d/1nVKsXiEXLviHOMY1iQEoyacvN93_SYCQwsOFfaP-dxo/edit#heading=h.67pgv1awhai).

Complexity:
A few libraries had dependency conflicts which required resolution. I believe this may have been improved by upgrades to Scala 2.13.x but there were dependencies which don't support it yet.

The conflicts were:

- `scala-java8-compat` -> Play used 0.9.1 and Play WS was using 1.x resulting in dependency incompatibility compile error. This was overcome by using an older version of Play WS (2.8.9) which was still using 0.9.1.
- `jackson-databind` -> AWS libraries and Play were using different version of `jackson-databind`. This was resulting in a runtime error which was overcome via [dependencyOverrides](https://www.scala-sbt.org/1.x/docs/Library-Management.html#Overriding+a+version) SBT feature.

## How to test

We should be able to test this works be deploying the changes to CODE or running locally. This should be a no-op from a user perspective. We would also expect areas such as logging to continue as before.

`workflow-frontend` is on the latest security patched versions of core dependencies.

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -50,7 +50,7 @@ class Application(
   }
 
   def index = AuthAction { request =>
-    Redirect(routes.Application.dashboard())
+    Redirect(routes.Application.dashboard)
   }
 
   def dashboard = app("Workflow")
@@ -80,7 +80,7 @@ class Application(
   def updateEditorialSupport = AuthAction(parse.form(EditorialSupportStaff.form)) { implicit request =>
     editorialSupportTeams.updateStaff(request.body)
     // Get the browser to reload the page once we've sucessfully updated
-    Redirect(routes.Application.editorialSupport())
+    Redirect(routes.Application.editorialSupport)
   }
 
   // limited tag fields we want output into the DOM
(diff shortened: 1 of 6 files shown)


jonathonherbert:
Tested in CODE by adding a stub article and modifying a few fields. Works as expected 🎉

jonathonherbert (approved):

rhystmills:
Merging now.

Had to update the conf in s3 - changing `application.secret` to `play.http.secret.key`

PR #368, opened 2022-06-27
Integrate repo with snyk GHA

NovemberTang:
This PR reliably integrates the repository with the snyk GitHub action. This PR has only been raised on repos that have already been tested to make sure scanning will work out of the box. ‘reliably integrated’ means that this action compares the hash of the last commit on main to the one that snyk has, and makes sure that they match.

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -1,12 +1,11 @@
+# This action runs snyk monitor on every push to main
 name: Snyk
 
 on:
-    schedule:
-        - cron: "0 6 * * *"
-    push:
-        branches:
-            - main
-    workflow_dispatch:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
 
 jobs:
   security:
@@ -14,5 +13,6 @@ jobs:
     with:
       DEBUG: true
       ORG: guardian
+      SKIP_NODE: false
     secrets:
        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}


.github/workflows/snyk.yml
```diff
@@ -1,18 +1,18 @@
+# This action runs snyk monitor on every push to main
 name: Snyk
 
 on:
-    schedule:
-        - cron: "0 6 * * *"
```
davidfurey, line 5: Apologies if this is obvious, but why is the cron no longer required?

NovemberTang: no worries! 
```
  push:
    branches:
      - main
```

already runs on every code change to main, so there's not much point also running it at 6am, its just duplicating runs.

NovemberTang: a slightly fuller explanation @davidfurey :
we're running `snyk monitor`, which just generates a list of all the dependencies and their dependencies and the dependencies dependencies etc and sends them off to snyk. If we ran that every day, most days there would be no changes, so we're only running it for every push.

Snyk then takes that list and runs it against its list of known vulnerabilities every day for us, and will notify the org members of any new exploits or vulnerabilities that come up. So as it will already have the most up to date dependency tree, there's no need to keep sending it every day.

davidfurey: Makes sense.

Thanks.

davidfurey (approved):

PR #376, opened 2022-09-23
Allow some endpoints to be authed using play-hmac

davidfurey:
## What does this change?

This is starting to address the comment in SharedSecretAuth "NB: we should also migrate to panda-hmac and SHA-256".

It doesn't actually stop us using the old sharedsecretauth, but it does mean a new user of this api can go straight to the future.

## How to test

* Deploy to code
* Check in browser that requests to `/api/content` and `/api/content/:composerId` do not encounter auth errrors

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -1,5 +1,6 @@
 package controllers
 
+import com.gu.pandahmac.HMACAuthActions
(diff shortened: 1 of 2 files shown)


andrew-nowak (approved):
changes seem reasonable and I continue to be authed to workflow code; lgtm

PR #378, opened 2022-10-24
Fix reauth for API endpoints

andrew-nowak:
## What does this change?

- Use APIHMAC actions for API endpoints
  - The two types of auth actions ('standard' vs API) behave differently when the user is unauthed or the auth has expired. 'standard' actions will return a redirect to the Google OAuth endpoint (useful if you're navigating, you get immediately pushed through the authentication flow and can interact with Google if necessary). The API actions will instead return a 4xx, which the app's frontend can detect and run an 'invisible' reauthentication flow.
  - When migrating a couple of the API endpoints to optionally use panda-HMAC auth, we accidentally picked the 'standard' action, which means that when the user is unauthed the action will attempt to redirect the XHR directly to Google. This will not succeed as the Google OAuth endpoint does not set CORS headers.
  - So, replace the HMAC actions with APIHMAC actions.

- While I'm here and not using nvm, this PR adds support for setting up with different node managers

## How to test

Open workflow, wait for initial page and api requests to load. Delete you panda cookie, then wait for the JS to poll the API. Before this change, reauth will fail and you will get a red banner saying unexpected error occurred. After this change, reauth succeeds.

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -59,15 +59,15 @@ class Api(
     }
   }
 
-  def content = HMACAuthAction.async(getContentBlock)
+  def content = APIHMACAuthAction.async(getContentBlock)
 
   def getContentByComposerId(composerId: String) = {
-      HMACAuthAction.async { implicit request =>
-        ApiResponseFt[Option[Stub]](for {
-          item <- getResponse(stubsApi.getStubByComposerId(stubDecorator, composerId))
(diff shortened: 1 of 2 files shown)


twrichards (approved):
makes total sense and since it affects the main `/api/content` endpoint I suspect this was the cause of much of the red banners seen by users lately

✅  TESTED in `CODE`

PR #381, opened 2022-10-28
use `bucketSsmLookup` in riff-raff.yaml 

twrichards:
remove `bucket` property from riff-raff.yaml - so it uses `bucketSsmLookup: true` by default (as per https://github.com/guardian/riff-raff/pull/884 and https://github.com/guardian/riff-raff/pull/885)

### DIFF
--- a/conf/riff-raff.yaml
+++ b/conf/riff-raff.yaml
@@ -4,7 +4,6 @@ deployments:
   workflow-frontend:
     type: autoscaling
     parameters:
-      bucket: workflow-dist
     dependencies: [workflow-frontend-ami-update, workflow-frontend-fluentbit]
   workflow-frontend-ami-update:
     type: ami-cloudformation-parameter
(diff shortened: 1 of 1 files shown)


conf/riff-raff.yaml
```diff
@@ -19,13 +18,11 @@ deployments:
...
     type: aws-lambda
     parameters:
-      bucket: workflow-dist
       prefixStack: false
       functionNames: [workflow-notification-]
       fileName: workflow-notification.zip
   workflow-frontend-fluentbit:
     type: aws-s3
     parameters:
-      bucket: workflow-dist
```
andrew-nowak, line 29: I don't think you can remove the bucket from aws-s3 (at least when I tried on grid I got failures when trying to preview the deployment, I can't remember if I actually tried to deploy it)

twrichards: oh that's sad, thanks for spotting - indeed the deploy failed in `CODE` - any plans to add that @akash1810? (I'm guessing the suggestion here would be to use `cdk-base` right?

twrichards: ok reinstated `bucket` for the `aws-s3` deploy step, and it deploys fine. ready for re-review @andrew-nowak - thank you

andrew-nowak (approved):

PR #382, opened 2022-10-28
expose the Panda HMAC header on request logging `X-Gu-Tools-Service-Name`

twrichards:
Following https://github.com/guardian/workflow-frontend/pull/376 we want to be able to associate request logging (see #380) with external services making use of Panda HMAC auth mechanism.


app/lib/LoggingFilter.scala
```diff
@@ -12,9 +13,9 @@ class LoggingFilter(override val mat: Materializer) extends Filter with Logging
     val ec = scala.concurrent.ExecutionContext.Implicits.global
 
     val startTime = System.currentTimeMillis
-    val headers = requestHeader.headers.getAll(LoggingContext.LOGGING_CONTEXT_HEADER)
-      .map(LoggingContext.fromHeader)
```
andrew-nowak, line 16: Are we okay not to pass the context header through this function any more? It looks like it's doing some meaningful parsing of the header

twrichards: fair point, re-instated

andrew-nowak (approved):

PR #384, opened 2022-11-10
Update readme on installation steps

aracho1:
Update README to make it more clear that  `api.url` in `/etc/gu/workflow-frontend.private.conf` needs to be amended as a common step for setting up both CODE and LOCAL configs + fix minor typos.


phillipbarron (approved):
LGTM 👍

PR #385, opened 2022-11-23
`Pinboard 📌` column

aracho1:
Co-authored-by: @twrichards 
Co-authored-by: @phillipbarron 

https://trello.com/c/s5n4JW0z/176-add-pinboard-visibility-in-a-new-workflow-column

**RELATED: https://github.com/guardian/pinboard/pull/219 - PREREQUISITE, MUST BE MERGED FIRST**

workflow_pinboard

This PR introduces a new column `Pinboard 📌`, which when active* displays a new column containing empty `<td>`s (defined in new file `public/components/content-list-item/templates/pinboard.html`) with class `content-list-item__field--pinboard` and importantly has the following `data` attributes...
- `data-pinboard-id`
- `data-working-title`
- `data-headline`

\* not active by default, but selectable from the column picker image IF the user has pinboard permission.

If this column is active then it loads the `pinboard.loader.js` script (just like [composer](https://github.com/guardian/flexible-content) and [grid](https://github.com/guardian/grid) do).

This gives pinboard all it needs to take over these elements using [React Portals](https://reactjs.org/docs/portals.html) after finding them via the class name, and extracting the `data` attributes - see https://github.com/guardian/pinboard/pull/219 for more detail.

We also add an empty div with id `pinboard-area` to the dashboard layout, to provide a place to dock the pinboard panel.

Unfortunately, because workflow-frontend has infinite scroll, it doesn't look possible to make the view sortable by the pinboard column (by unread messages for example) - but might revisit in future.

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -1,4 +1,5 @@
 import com.gu.pandomainauth.PanDomainAuthSettingsRefresher
+import com.gu.permissions.{PermissionsConfig, PermissionsProvider}
 import com.gu.workflow.api.{DesksAPI, SectionDeskMappingsAPI, SectionsAPI, StubAPI}
 import com.gu.workflow.lib.TagService
 import com.gu.workflow.util.AWS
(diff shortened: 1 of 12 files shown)


phillipbarron (dismissed):
Looks good to me 👍

pradasa:
The scrolling behaviour looks good to me!
I have found one non-critical bug: 
1) If you open a pinboard
2) Scroll down
3) Reach a row where the WT is longer than the rows above, the columns shift to the right to adjust the spacing. (The culprit in the screenshot is the row with WT: AF Auto Test: Working title 107559. As soon as it appears on screen, the columns shift)

Screenshot: Comparison of column spacing between before and after the culprit row appears
Frame 442

twrichards:
> The scrolling behaviour looks good to me! I have found one non-critical bug:
> 
> 1. If you open a pinboard
> 2. Scroll down
> 3. Reach a row where the WT is longer than the rows above, the columns shift to the right to adjust the spacing. (The culprit in the screenshot is the row with WT: AF Auto Test: Working title 107559. As soon as it appears on screen, the columns shift)

Hey @pradasa - thanks for the observation and clear comment - we've noticed the same during development, however this is workflows existing behaviour to do with its 'infinite scroll', i.e. it distributes the columns based on their contents (except some columns which have fixed widths), so as the user scrolls down and more rows are loaded into the table, it readjusts the column widths automatically (which is HTML/CSS/Browser behaviour, not special logic coded into workflow). 

If you turn the pinboard column off (but keep enough other columns turned on), you'll observe the same.

If you still consider this important, I think the only way to address would be to give all columns in workflow a fixed width, but I'll leave it with you as to whether that's desirable.

aracho1:
Looks good to me, just a minor UI bug
image

phillipbarron (dismissed):
Reviewed with Ara - looks good - there is the small UI issue Ara pointed out

phillipbarron (approved):

PR #393, opened 2023-01-02
Add rights editing to table

jonathonherbert:
## What does this change?

Adds the ability to filter by rights information in the content table, and set rights – either all of them via the column, or on a per-right basis in the 'management' drawer.

https://user-images.githubusercontent.com/7767575/210387748-f0a99d1c-ca66-404f-bafc-9295c2a10fed.mov

See https://github.com/guardian/workflow/pull/1101 for notes on how the data is stored.

## How to test

In CODE, try to:

- Add/remove rights via the table or management drawer
- Use the filter on the left hand side

Does this behave as expected? 

~NB: there's one bug! Because we go back to Composer to set this information, and then wait for Composer to report the new value via flexible-content-stream, sometimes workflow-frontend will update _before_ Composer has a chance to report the new value. Our client cannot distinguish between these sorts of updates, so if you click the button, it will occasionally revert to the old value before eventually displaying the new one:~

https://user-images.githubusercontent.com/7767575/210393319-386022b5-e52b-4967-943d-de6d4acf8aa6.mov

~There are a few ways I can imagine solving (you could pause state updates until the data reflected what you expected, for example), but I think it makes sense to get this out first.~

Above bug should be fixed by [eac696b](https://github.com/guardian/workflow-frontend/pull/393/commits/eac696b71717d63f56fb8fe35f9ceda86444d24f).

### DIFF
--- a/public/components/directives/ui-edit-rights.js
+++ b/public/components/directives/ui-edit-rights.js
@@ -30,17 +30,6 @@ export const uiEditRights = (wfComposerService) => ({
           </label>
         </div>
         <div>
-          <label class="ui-edit-rights__label" for="rights-subscription-databases">
-            <input type="checkbox"
-                   id="rights-subscription-databases"
-                   class="ui-edit-rights__input"
-                   ng-model="contentItem.item.rightsSubscriptionDatabases"
-                   ng-click="$event.stopPropagation()"
-                   ng-change="updateRights()" />
-            <span>Subscription databases</span>
-          </label>
-        </div>
-        <div>
           <label class="ui-edit-rights__label" for="rights-developer-community">
             <input type="checkbox"
                    id="rights-developer-community"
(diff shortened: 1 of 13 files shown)


public/components/directives/ui-edit-rights.js
```diff
@@ -73,10 +73,32 @@ export const uiEditRights = (wfComposerService) => ({
...
+     */
+    const lockRightsValuesForDuration = (shouldHaveRights) => {
+      const cancelWatch = scope.$watch("contentItem", contentItem => {
+        contentItem.item.rightsSyndicationAggregate = shouldHaveRights;
+        contentItem.item.rightsSubscriptionDatabases = shouldHaveRights;
+        contentItem.item.rightsDeveloperCommunity = shouldHaveRights;
+      });
+
+      // Lock rights data for one polling cycle.
+      setTimeout(cancelWatch, 5000);
```
alinaboghiu, line 92: What would happen if user tried to change the value twice in quick succession? Would the first update work but the second one not? And if latter, then would we want to grey out the boxes in the UI until they are available again, or display some kind of loading indicator?

jonathonherbert: It shouldn't be possible to change the value twice on a single piece of content with 'add all rights', as the button disappears as soon as it's used. ('Remove all rights' is no longer a thing 😢) 

There will be an odd synchronisation issue if the user _then_ opens the drawer and observes the behaviour of the syndication checkboxes, which will sometimes regress on the first update, before likely syncing up on the second.

To improve this, we could simultaneously write to Composer, and Workflow backend locally – workflow backend then anticipates the change, and does not update the client with incorrect information, but will update via the stream shortly.

jonathonherbert: (Alongside this, trying to figure out how much effort to expend given that the service is eventually consistent and we know the workflow may be somewhat temporary!)

Fweddi (approved):

PR #395, opened 2023-01-12
Add byline column

jonathonherbert:
## What does this change?

Adds a byline column to the content table, to the right of the headline column. It's not visible by default.

Screenshot 2023-01-12 at 17 25 17

## How to test

- Deploy both this PR and https://github.com/guardian/workflow/pull/1104 to CODE, and modify a few pieces with bylines, or add bylines to pieces tracked in workflow. Add the byline column to your table. Do the bylines appear as expected?

An easier job for the Rights Ops team as they add rights to content.


alinaboghiu (approved):
Looks great!

PR #398, opened 2023-02-13
replace Google API with permissions cache

twrichards:
https://trello.com/c/eTU2qkTG/1077-update-google-auth

Co-authored-by: @andrew-nowak 

### Background 
In workflow UI it's long been possible to 'assign' items to other people, which relied on Google APIs for populating the list of users (autocomplete), which looked like this...

| Autocomplete | Assigned |
| --- | --- |
| image | image | 

Unfortunately our current mechanism of auth with these Google APIs is coming to an end soon...
image
... in https://github.com/guardian/workflow-frontend/pull/397 we attempted to upgrade the auth approach, but this introduces some invasive/frequent pop-ups and requires storing tokens to local storage 😬.

Also worth noting that over the years a number of users have occasional trouble with the blue banner which prompts users to auth with Google, for which CP have documented workarounds - not ideal.

### Solution
Since we have a full list of users in our permissions cache, we decided to expose this (see https://github.com/guardian/permissions/pull/166)  and use it to populate the autocomplete (via a new `/api/people?prefix=` endpoint) and then store only the assignee email.

workflow_assigning

This allowed us to remove the dependency on Google entirely and all the UI complexity associated. **This PR represents a reasonable simplification overall 🎉 .**


rebecca-thompson (approved):
Fantastic PR description, thanks!

I have tested this feature in CODE and it works as expected. 

Code looks good 👍

Shame there are not more unit tests in this project as it would have been good to have tests around this feature

PR #401, opened 2023-06-01
Replace deprecated, ARM incompatible node-sass library

Fweddi:
## What does this change?

`node-sass` has been deprecated, and it doesn't seem any versions of it are compatible with `ARM` (now the architecture of most dev machines). This PR replaces the `node-sass` library with `sass`, which requires a more recent node version.

The new library requires a couple of syntax changes, which can be found in this PR.

## How to test

With the right Janus permissions (workflow and **capi**):

1. Run the setup script
2. Run the start script
3. Visit https://workflow.local.dev-gutools.co.uk/dashboard and see it render correctly

Alternatively, it's worth testing this on CODE - where Presence etc. is working. Here any style changes will be easier to spot.


jonathonherbert (approved):
Deployed to CODE, looks good!

PR #408, opened 2023-10-11
proxy `preferences` calls via `workflow-frontend` 

twrichards:
...to benefit from CloudFront which is in front of workflow-frontend but not editorial-preferences (mainly to benefit users in AUS).

Related: #407 

BONUS: taken this opportunity to redirect calls to preferences.local to preferences.code when https://github.com/guardian/editorial-preferences is NOT running locally.

## TESTING

- LOCALLY; once running (with the tunnel, as detailed in README) notice your code workflow column choices etc. are use and that if you update* then refresh in CODE workflow you'll see the same choices
- CODE; reading and updating* column configuration etc. works as before (you can observe the preferences related Network calls using Developer Tools - they should be using the workflow domain, not preferences)

\* column configuration can be changed by scrolling to the very right hand side, clicking image at the end of the column headings, changing some columns, scrolling to the bottom of the column list and clicking image

### DIFF
--- a/app/config/Config.scala
+++ b/app/config/Config.scala
@@ -69,11 +69,12 @@ class Config(playConfig: Configuration) extends AwsInstanceTags with Logging {
   lazy val presenceUrl: String = s"wss://presence.$domain/socket"
   lazy val presenceClientLib: String = s"https://presence.$domain/client/1/lib.js"
 
-  lazy val preferencesUrl: String = s"https://preferences.$domain/preferences"
+  lazy val preferencesHost: String = s"preferences.$domain"
+  lazy val preferencesUrl: String = s"https://$preferencesHost/preferences"
 
-  lazy val pinboardLoaderUrl: String = s"https://pinboard.${stage.appDomain}/pinboard.loader.js"
+  lazy val pinboardLoaderUrl: String = s"https://pinboard.$domain/pinboard.loader.js"
 
-  lazy val tagManagerUrl: String = s"https://tagmanager.${stage.appDomain}"
+  lazy val tagManagerUrl: String = s"https://tagmanager.$domain"
 
   lazy val capiPreviewIamUrl: String = playConfig.get[String]("capi.preview.iamUrl")
   lazy val capiPreviewRole: String = playConfig.get[String]("capi.preview.role")
--- a/app/config/Config.scala
+++ b/app/config/Config.scala
@@ -69,7 +69,7 @@ class Config(playConfig: Configuration) extends AwsInstanceTags with Logging {
   lazy val presenceUrl: String = s"wss://presence.$domain/socket"
   lazy val presenceClientLib: String = s"https://presence.$domain/client/1/lib.js"
 
-  lazy val preferencesUrl: String = s"https://preferences.${stage.appDomain}/preferences"
+  lazy val preferencesUrl: String = s"https://preferences.$domain/preferences"
 
   lazy val pinboardLoaderUrl: String = s"https://pinboard.${stage.appDomain}/pinboard.loader.js"
 
--- a/app/controllers/PreferencesProxy.scala
+++ b/app/controllers/PreferencesProxy.scala
@@ -4,7 +4,7 @@ import com.gu.pandomainauth.PanDomainAuthSettingsRefresher
 import com.gu.workflow.util.{Code, Dev}
 import config.Config
 import play.api.Logging
-import play.api.libs.ws.WSClient
+import play.api.libs.ws.{EmptyBody, InMemoryBody, WSClient}
 import play.api.mvc.{BaseController, ControllerComponents}
 
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -17,13 +17,20 @@ class PreferencesProxy(
   override val panDomainSettings: PanDomainAuthSettingsRefresher
 ) extends BaseController with PanDomainAuthActions with Logging {
 
-  private def proxyRequest(relativePath: String) = APIAuthAction.async { request =>
+  private def proxyRequest(relativePath: String) = APIAuthAction(parse.byteString).async { request =>
     val url = s"${config.preferencesUrl}/$relativePath"
     val requestHeaders = request.headers.toSimpleMap + (
       "Host" -> config.preferencesHost
     )
+    val requestBody = if (request.hasBody) {
+      InMemoryBody(request.body)
+    } else {
+      EmptyBody
+    }
     wsClient.url(url)
       .withHttpHeaders(requestHeaders.toSeq: _*)
+      .withBody(requestBody)
+      .withFollowRedirects(false) // ensure browser handles all redirects NOT this proxy
       .execute(request.method)
       .map { response =>
         val responseHeaders = response.headers.mapValues(_.head) + (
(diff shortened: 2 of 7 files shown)


app/controllers/PreferencesProxy.scala
```diff
@@ -0,0 +1,40 @@
...
+      .withHttpHeaders(request.headers.toSimpleMap.toSeq: _*)
+      .execute(request.method)
+      .map { response =>
+        Ok(response.body).withHeaders(response.headers.mapValues(_.head).toSeq: _*)
+      }
+      .recover { case e =>
+        logger.error(s"Error proxying request to $url", e)
+        TemporaryRedirect( // this ensures PUT doesn't get transformed to GET, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#temporary_redirections
+          if (config.isDev) url.replaceAll(Dev.appDomain, Code.appDomain)
+          else url
```
davidfurey, line 32: What kind of failure are you anticipating where the proxy cannot connect to preferences but the client can?

app/controllers/PreferencesProxy.scala
```diff
@@ -0,0 +1,40 @@
...
+  override val config: Config,
+  override val controllerComponents: ControllerComponents,
+  override val wsClient: WSClient,
+  override val panDomainSettings: PanDomainAuthSettingsRefresher
+) extends BaseController with PanDomainAuthActions with Logging {
+
+  private def proxyRequest(relativePath: String) = APIAuthAction.async { request =>
+    val url = s"${config.preferencesUrl}/$relativePath"
+    wsClient.url(url)
+      .withHttpHeaders(request.headers.toSimpleMap.toSeq: _*)
```
davidfurey, line 23: Will this pass through the original host header (for workflow) rather than the new host header (for preferences)?

If so, you might find that this works now but stops working when a seemingly innocuous change is made to how presence is served.

app/controllers/PreferencesProxy.scala
```diff
@@ -0,0 +1,40 @@
...
+      .execute(request.method)
+      .map { response =>
+        Ok(response.body).withHeaders(response.headers.mapValues(_.head).toSeq: _*)
```
davidfurey, line 26: I think you should also handle passing on the HTTP status code to the client

app/config/Config.scala
```diff
@@ -69,7 +69,7 @@ class Config(playConfig: Configuration) extends AwsInstanceTags with Logging {
   lazy val presenceUrl: String = s"wss://presence.$domain/socket"
   lazy val presenceClientLib: String = s"https://presence.$domain/client/1/lib.js"
 
-  lazy val preferencesUrl: String = s"https://preferences.${stage.appDomain}/preferences"
+  lazy val preferencesUrl: String = s"https://preferences.$domain/preferences"
```
davidfurey, line 72: It isn't immediately obvious from the diff that this is a noop.  But since you are changing this, why not also change line 74?

davidfurey (commented):
Have you checked what cache control headers preferences returns, and if it doesn't return any cache-control headers, what cloudfront's default caching behaviour is?

app/controllers/PreferencesProxy.scala
twrichards: yeah that was silly, I've refactored now such that the isDev case is handled in the match and otherwise serves `InternalServerError` on failure

app/controllers/PreferencesProxy.scala
twrichards: great spot - I've made the `Host` header explicit now

app/controllers/PreferencesProxy.scala
twrichards: absolutely - done in https://github.com/guardian/workflow-frontend/pull/408/commits/85689dfa9ae4ddbb2a223a35ca4a27c604d675f9

app/config/Config.scala
twrichards: I've not made it any more obvious that it's a noop, but I have made it consistent across this whole file now (i.e. used `$domain` everywhere as more succinct

twrichards:
> Have you checked what cache control headers preferences returns, and if it doesn't return any cache-control headers, what cloudfront's default caching behaviour is?

ooo excellent point, I haven't checked Cloudfront's cacheing, but I've added an explicit cache control header (to never cache) in https://github.com/guardian/workflow-frontend/pull/408/commits/85689dfa9ae4ddbb2a223a35ca4a27c604d675f9

app/controllers/PreferencesProxy.scala
```diff
@@ -0,0 +1,47 @@
...
+    wsClient.url(url)
+      .withHttpHeaders(headers.toSeq: _*)
+      .execute(request.method)
+      .map { response =>
+        new Status(response.status)(response.body).withHeaders(response.headers.mapValues(_.head).toSeq: _*)
+      }
+      .recover {
+        case _ if config.isDev =>
+          TemporaryRedirect( // this ensures PUT doesn't get transformed to GET, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#temporary_redirections
+            if (config.isDev) url.replaceAll(Dev.appDomain, Code.appDomain)
```
davidfurey, line 35: You've already checked for isDev in the pattern match, so don't need it here as well

app/controllers/PreferencesProxy.scala
```diff
@@ -0,0 +1,47 @@
...
+  override val controllerComponents: ControllerComponents,
+  override val wsClient: WSClient,
+  override val panDomainSettings: PanDomainAuthSettingsRefresher
+) extends BaseController with PanDomainAuthActions with Logging {
+
+  private def proxyRequest(relativePath: String) = APIAuthAction.async { request =>
+    val url = s"${config.preferencesUrl}/$relativePath"
+    val headers = request.headers.toSimpleMap + (
+      "Host" -> config.preferencesHost,
+      "Cache-Control" -> "private, no-cache, no-store, must-revalidate, max-age=0", // do not cache whatsoever
```
davidfurey, line 24: This cache-control header should be in the response on line 30, not in the request that you are sending to the preferences service.

davidfurey (commented):
Thanks for the changes so far, just a couple more needed

app/controllers/PreferencesProxy.scala
twrichards: oops, missed that - fixed in https://github.com/guardian/workflow-frontend/pull/408/commits/9641702b301bc220da763a3dc3d791b9836fa5fc

app/controllers/PreferencesProxy.scala
twrichards: 🤦  - thanks for catching that - updated in https://github.com/guardian/workflow-frontend/pull/408/commits/9641702b301bc220da763a3dc3d791b9836fa5fc - plus better variable name(s)

davidfurey (dismissed):
lgtm

twrichards:
gahhh, after doing some final testing, I see the PUTs just `504` 
image
...on for Monday 😢

twrichards:
> gahhh, after doing some final testing, I see the PUTs just `504` image ...one for Monday 😢

Fixed on the train back from footy. Went for the `ByteString` approach borrowed from editorial-wires.

phillipbarron (approved):
LGTM

PR #412, opened 2023-11-21
allow recipes in DEV to call Workflow CODE

twrichards:
In order to facilitate [Pinboard 📌 in guardian/recipes](https://github.com/guardian/recipes/pull/65) we want to allow recipes to make AJAX calls to workflow (since Pinboard requires things to be tracked in Workflow) which is currently blocked by CORS.

This PR allows such calls but also lets LOCAL recipes to use workflow CODE (similar to MAM).

### DIFF
--- a/app/config/Config.scala
+++ b/app/config/Config.scala
@@ -42,14 +42,9 @@ class Config(playConfig: Configuration) extends AwsInstanceTags with Logging {
     case _ => Set(composerUrl, s"https://composer-secondary.$domain")
   }
 
-  private lazy val mediaAtomMakerUrls: Set[String] = stage match {
-    case Code => Set(mediaAtomMakerUrl, s"https://video.${Dev.appDomain}") // allow MAM in DEV to call Workflow CODE
-    case _ => Set(mediaAtomMakerUrl)
-  }
-
-  private lazy val atomWorkshopUrls: Set[String] = stage match {
-    case Code => Set(s"https://atomworkshop.${Dev.appDomain}", atomWorkshopUrl) // allow MAM in DEV to call Workflow CODE
-    case _ => Set(atomWorkshopUrl)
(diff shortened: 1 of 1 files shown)


app/config/Config.scala
```diff
@@ -37,6 +37,8 @@ class Config(playConfig: Configuration) extends AwsInstanceTags with Logging {
   lazy val mediaAtomMakerUrl: String = s"https://video.$domain"
   lazy val atomWorkshopUrl: String = s"https://atomworkshop.$domain"
 
+  private lazy val recipesUrl: String = s"https://recipes.${if(isProd) Prod.appDomain else Dev.appDomain}"
```
davidfurey, line 40: Is this different from:
```suggestion
  private lazy val recipesUrl: String = s"https://recipes.$domain"
```

app/config/Config.scala
```diff
@@ -52,6 +54,11 @@ class Config(playConfig: Configuration) extends AwsInstanceTags with Logging {
     case _ => Set(atomWorkshopUrl)
   }
 
+  private lazy val recipesUrls: Set[String] = stage match {
+    case Code => Set(s"https://recipes.${Dev.appDomain}", recipesUrl) // allow recipes in DEV to call Workflow CODE
+    case _ => Set(recipesUrl)
+  }
+
```
davidfurey, line 61: Now that we are doing this three times, it might be clearer to refactor into a function?

If not, while you are here can you correct the comment for `atomWorkshopUrls` which says that it allows **MAM** in DEV to call workflow CODE?

davidfurey (commented):
Change seems reasonable to me.

app/config/Config.scala
twrichards: good spot, PR refactored now

davidfurey (dismissed):

davidfurey (approved):

PR #414, opened 2023-11-27
fix admin navbar position (was obscuring stuff beneath)

twrichards:
Noticed the tabs were obscured when trying to add a Section. Looks like there was a previous attempt at this which wasn't working but also padding content down by an arbitrary amount to accommodate navbar is not great, so I've removed in favour of a `position: sticky` directly on the element, so it can be seen alongside the `navbar-fixed-top` Bootstrap class, since its changing that behaviour (Bootstrap doesn't have a `navbar-sticky` or `navbar-sticky-top`).

### Before
image


simonbyford (approved):
Makes sense 👍

PR #415, opened 2024-01-02
Enable key takeaway article creation

Georges-GNM:
## What does this change?

Following on the work in [composer](https://github.com/guardian/flexible-content/pull/4490); workflow has also been identified as an entry point to creating key takeaway articles, so we're adding this as an easy entry point. 

This also hides this logic behind a rudimentary and specific (as in not abstracted and usable for other purposes) feature switch (via the userPreference service, so it should remain once activated), the toggle for which is behind shift + F12

## How to test

Load the branch, ideally in code to verify the composer integration. 

Initially nothing should be changed in the "Create new" dropdown. Pressing shift + f12 should bring up a small button below the user name: 
image

Clicking on it should force the page to reload, and then, clicking on the "create new" dropdown should show this: 
image

Clicking on anything other than "article" or "key takeaway" should show no change; either of those should add a "Format" option at the top: 
image

Creating a new article with "standard" should show a regular article in the composer file; however if you select the "key takeaway" option when creating a new article, this should result in a composer file with the "key takeaway" option/display hint set. 

You should also be able to switch between the two formats in the dropdown, and the composer file gets created with the correct displayhint. 

## Quick thought 

It may be useful to add some telemetry here when also tackling this for composer.


public/components/feature-switch/feature-switch.js
```diff
@@ -0,0 +1,35 @@
...
+function wfFeatureSwitchController ($scope, wfPreferencesService) {
+
+    function checkPreference(){
+    wfPreferencesService.getPreference('featureSwitch').then((data) => { $scope.featureSwitch = data;})
+    }
+
+    checkPreference();
+
+    $scope.toggleFeatureSwitch = () => {
+        let newValue = !$scope.featureSwitch;
```
twrichards, line 32:
```suggestion
        const newValue = !$scope.featureSwitch;
```

public/components/feature-switch/feature-switch.html
```diff
@@ -0,0 +1,7 @@
+<button class="feature-switch-toggle" ng-click="toggleFeatureSwitch()" title="Toggle key takeaways creation">
+  Toggle key takeaways creation
+  <div
+    class="feature-switch-toggle__icon"
+    ng-class="{'feature-switch-toggle__icon--active': featureSwitch.visible }"
+  ></div>
+</button>
```
twrichards, line 7: this could be modelled as a checkbox with the label as just `Key Takeaways creation`

or failing that, you could at least alter the text based on current value, something like this
```suggestion
<button class="feature-switch-toggle" ng-click="toggleFeatureSwitch()" title="Toggle key takeaways creation">
 {featureSwitch ? 'ENABLE' : 'DISABLE'} Key Takeaways creation
  <div
    class="feature-switch-toggle__icon"
    ng-class="{'feature-switch-toggle__icon--active': featureSwitch.visible }"
  ></div>
</button>
```

public/components/stub-modal/stub-modal.js
```diff
@@ -252,6 +262,7 @@ function StubModalInstanceCtrl($rootScope, $scope, $modalInstance, $window, conf
 
     $scope.ok = function (addToComposer, addToAtomEditor) {
         const stub = $scope.stub;
+        console.log("stub", stub)
```
twrichards, line 265:
```suggestion
```

rhystmills:
A good option for telemetry is https://github.com/guardian/grid/pull/3974, which uses https://github.com/guardian/editorial-tools-user-telemetry-service - I did a fairly minimal integration for the Grid/Kahuna in this PR: https://github.com/guardian/grid/pull/3974

That could be added in a follow-up PR,.

(I noticed workflow has Google Analytics, which we don't use any more, presumably that is where user measurements went in the past)

Georges-GNM:
> A good option for telemetry is [guardian/grid#3974](https://github.com/guardian/grid/pull/3974), which uses [guardian/editorial-tools-user-telemetry-service](https://github.com/guardian/editorial-tools-user-telemetry-service) - I did a fairly minimal integration for the Grid/Kahuna in this PR: [guardian/grid#3974](https://github.com/guardian/grid/pull/3974)
> 
> That could be added in a follow-up PR,.
> 
> (I noticed workflow has Google Analytics, which we don't use any more, presumably that is where user measurements went in the past)

Would probably be best removing GA at some point then. 

Yeah, was thinking of that telemetry service; also reckon it makes sense to leave for a follow up PR once we've identified what we want to be capturing and doing in composer.

public/components/feature-switch/feature-switch.js
```diff
@@ -0,0 +1,35 @@
...
+    };
+}
+
+function wfFeatureSwitchController ($scope, wfPreferencesService) {
+
+    function checkPreference(){
+    wfPreferencesService.getPreference('featureSwitch').then((data) => { $scope.featureSwitch = data;})
+    }
+
+    checkPreference();
```
twrichards, line 29: if we're not re-using this function, we could simplify these lines
```suggestion
    wfPreferencesService.getPreference('featureSwitch').then((data) => { $scope.featureSwitch = data;})
```

public/lib/content-service.js
```diff
@@ -23,8 +24,25 @@ angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService
...
+                        "liveblog": "Live blog",
+                        "gallery": "Gallery",
+                        "interactive": "Interactive",
+                        "picture": "Picture",
+                        "audio": "Audio",
+                        "atom": "Video/Atom"
+                    })} 
+
+                getTypes() {
+                    return wfPreferencesService.getPreference('featureSwitch').then((isSwitchActive) => {
```
twrichards, line 43: since you've abstracted the feature switch logic into its own service (nice one btw 👏 ) might be worth using that here to keep everything nicely encapsulated (albeit with a function exposed to retrieve the feature switch value)

twrichards (approved):
tested this in `CODE` earlier, works perfectly - amazing work 👏 

I've left various comments, mostly minor. The big one is about representing multiple feature switches, wouldn't be much work and would probably be worth it to avoid data building up for users and having to map that later if we want more feature switches later.

Approving - so you're unblocked for merging in time for demo - but the above is worth it IMO

PR #417, opened 2024-01-16
Add logging of push notifications to track usage

davidfurey:
## What does this change?

Logs usage of push notifications feature so that we can decide whether to upgrade the dependencies required for this feature, or remove the feature.


andrew-nowak:
seems okay, but I think this would be already covered in line 52... but also this runs in a lambda so you could see some usage from the monitoring graphs on the lambda console.
(separately, it looks like logging needs some work on that lambda, the only non-lambda-runtime logs are the good ol
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
```
)

davidfurey:
Ah, thanks for pointing that out. I was focussing on `SubscriptionsAPI.scala`, it was a last-minute move into `Notifier.scala` so I missed that the logging already existed. I'll close this and open a PR to fix the lambda logging instead

PR #418, opened 2024-01-16
Remove akka agent

davidfurey:
## What does this change?

Akka-agent was removed from Akka years ago. Let's stop using it in workflow-frontend.

Conveniently the only place we were using it was to wrap a constant value, so it was easy to remove.


andrew-nowak (approved):
👏

PR #419, opened 2024-01-17
Fix lambda logging

davidfurey:
## What does this change?

Adds a dependency on slf4j-simple to (hopefully) fix the notifier lambda logging.

See https://www.slf4j.org/codes.html#StaticLoggerBinder

## How to test

Deploy to CODE. Check that notifier logs appear in central elk


davidfurey:
Tested on CODE - more interesting logging is now coming through

andrew-nowak (approved):

PR #420, opened 2024-01-17
Upgrade Scala/Play/SBT

davidfurey:
## What does this change?

Upgrades to Java 11, Scala 2.13, SBT 1.9.6 and Play 2.9.1

Also swaps out scalacache-caffeine for scaffeine. The API for scalacache-caffeine had changed quite a bit with the bump to a Scala 2.13 compatible version, and it only has milestone releases for v1.0. In contrast, Scaffeine seems better supported and is used in other projects

## How to test

Deploy to CODE, check app behaves as expected.

No one notices, but Snyk vulnerabilities drop

### DIFF
--- a/app/controllers/Admin.scala
+++ b/app/controllers/Admin.scala
@@ -134,7 +134,7 @@ class Admin(
   )
 
   def assignSectionToDesk = (AuthAction andThen PermissionFilter).async { implicit request =>
-    assignSectionToDeskForm.bindFromRequest.fold(
+    assignSectionToDeskForm.bindFromRequest().fold(
       formWithErrors => Future(BadRequest("failed to update section assignments")),
       sectionAssignment => {
         sectionDeskMappingsAPI.assignSectionsToDesk(sectionAssignment.desk, sectionAssignment.sections.map(id => id.toLong))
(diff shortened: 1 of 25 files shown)


davidfurey:
Deployed to CODE, seems to work, but perhaps warrants more extensive testing

Georges-GNM (approved):
Branch is on code and seems to be working as intended. 

Looks like scalacache was being used for something related to tags which, from my understanding of them in workflow, seem to be behaving as normal, so looks good to me 👍

app/controllers/Admin.scala
```diff
@@ -250,7 +250,7 @@ class Admin(
       tagIds <- tagIdsFuture
       selectedSectionOption <- selectedSectionOptionFt
     } yield {
-      val capiKeyJson: Json = parser.parse(s"""{"CAPI_API_KEY": ${config.capiKey}}""").right.get
+      val capiKeyJson: Json = parser.parse(s"""{"CAPI_API_KEY": ${config.capiKey}}""").toOption.get
```
bryophyta, line 253: @davidfurey this line seems to be throwing an error in CODE (I don't have Workflow admin permissions for PROD so can't check it there). It looks like this was the last commit to touch this code (though seems likely that the bug preceded this commit) -- do you have any context on what the code here's doing? Happy to try and raise a PR if it'd be useful.

bryophyta: To reproduce:
https://workflow.code.dev-gutools.co.uk/admin/sections-and-tags

Logs: https://logs.gutools.co.uk/s/editorial-tools/app/discover#/doc/664d234d-de11-5f34-8ac5-9ee2089ac473/logstash-ed-tools-workflow-2024.11.21?id=cZ1JT5MBT0J6efU8a76d

PR #421, opened 2024-01-18
Fix create new dropdown 

Georges-GNM:
## What does this change?

Fixes a bug introduced by #415 where the request for the user's featureSwitch preference was causing the dropdown to not show any contentTypes if that preference hadn't been set previously.

### DIFF
--- a/public/lib/content-service.js
+++ b/public/lib/content-service.js
@@ -40,8 +40,10 @@ angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService
                     })} 
 
                 getTypes() {
-                    return wfPreferencesService.getPreference('featureSwitch').then((isSwitchActive) => {
+                    return wfPreferencesService.getPreference('featureSwitch')
+                    .then((isSwitchActive) => {
                         return isSwitchActive === true ? this.provideStandardAndNewFormats() : this.provideStandardFormats()})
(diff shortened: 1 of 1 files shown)


rhystmills:
Working as intended for me in CODE with preferences set or unset 👍

rhystmills (approved):

public/lib/content-service.js
```diff
@@ -40,8 +40,10 @@ angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService
                     })} 
 
                 getTypes() {
-                    return wfPreferencesService.getPreference('featureSwitch').then((isSwitchActive) => {
+                    return wfPreferencesService.getPreference('featureSwitch')
+                    .then((isSwitchActive) => {
                         return isSwitchActive === true ? this.provideStandardAndNewFormats() : this.provideStandardFormats()})
+                    .catch((err) => {return this.provideStandardFormats()})
```
twrichards, line 46: Nice fix, but why would it error?

Georges-GNM: From quickly looking at the getPreference() function in `preferences-service.js`, it looks like it's meant to return a Promise.reject() if it doesn't find a preference set for a particular value - so I guess that needed to be handled.

PR #422, opened 2024-01-18
Hide format dropdown behind switch

Georges-GNM:
## What does this change?

This bit of logic was absent from #415 but still necessary at this stage as we don't want the modal to show format variations yet. 

## Testing

On code, if the key takeaways feature switch isn't activated, the modal should not display a format dropdown. If it is, it should only when creating an article or key takeaway.


rhystmills (approved):
Working as intended in CODE, with the format dropdown not showing if the feature switch is off 👍 

I have noticed - if the user selects 'Article' then changes the format via the dropdown, they are still able to select a template, which seems to override the key takeaways display hint in Composer. 

I'd be inclined to have Article and Key Takeaways to be completely separate options and drop the format dropdown - it feels a bit clearer to me and more in line with what we now have in Composer. This would also avoid us needing to address the above issue.

PR #427, opened 2024-03-28
Replace single feature switch with multi-switchboard and add Q&A switch

rhystmills:
## What does this change?

This PR replaces our single feature switch with a multi-switch switchboard, and adds an additional switch for the Q and A format. As before, this is available behind the shortcut <kbd>shift-f12</kbd> (<kbd>shift-fn-f12</kbd> for me on Mac).

image

The switches are still stored in preferences, meaning they end up in a plain text entry in DynamoDB. However, they're now a  JSON value under the key `featureSwitches` rather than a single boolean under `featureSwitch` - along the lines of:

`{"keyTakeawaysSwitch":false,"qAndASwitch":true}`

A user won't have a value stored until they update one of the feature switches (or we add a row to DynamoDB directly).

## Note before merging:

<em>**Before merging this PR,**</em> we may want to manually re-add entries to DynamoDB for users who have opted into the old feature switch format, using the new format, to ensure continuity of access to Key Takeaways.

## How to test

1. Run workflow-frontend locally according to the instructions in the readme, or deploy to CODE. (Local uses CODE preferences so you can still access preferences there).
2. Click the page and hit <kbd>shift-f12</kbd>. Does a feature switch board appear in the top right?
3. Change one of the preferences. Does the page refresh briefly afterwards?
4. Look in the 'Create new' menu. Has the format you changed honoured your changed setting?

## Accessibility

The feature switch panel is not keyboard accessible - as it's used infrequently by a small number of users I think this is acceptable. We can manually add permissions for affected users in DynamoDB if we need to.


twrichards (commented):
tested on `CODE` and although it was setting the preferences correctly... 
image
... I was getting the following...
image
... when trying to actually create the articles, getting `400 Bad Request`...
image

twrichards:
is there a reason to store the JSON as a String?
DynamoDB supports `Map` (effectively JSON object) and looks like this...
image
... its also queryable too IIRC

Fweddi:
> tested on `CODE` and although it was setting the preferences correctly... image ... I was getting the following... image ... when trying to actually create the articles, getting `400 Bad Request`... image

This should now be fixed by [`aab8462` (#427)](https://github.com/guardian/workflow-frontend/pull/427/commits/aab846289068e90c73bed4aa6800d47b7f2b6730)

rhystmills:
> is there a reason to store the JSON as a String? DynamoDB supports `Map` (effectively JSON object) and looks like this... image ... its also queryable too IIRC

I wasn't aware of that - [preferences](https://github.com/guardian/editorial-preferences) seems to have the assumption baked in that the values will be strings "The preference itself is stored as a string, which is commonly serialised data, e.g. JSON."

Fweddi (approved):

PR #433, opened 2024-05-08
dependency upgrade

dblatcher:
## What does this change?

Applies some dependency upgrades and a minor upgrade to node to reduce vulnerabilities.

This PR is limited to "easy" changes that didn't require changes to the source code or config. There are remaining issues to be addressed on future PR's. In particular:
 - upgrading Mocha to v6 caused the test suite to fail
 - upgrading to the earliest version of angular with a fix caused the runtime errors (project would run but the dashboard UI failed)
 - fixing issue with karma-webpack would involve changing to webpack 5

Some of the remaining issues are from Angular, with no patches available at the latest version.

## How to test

Project will still build, pass tests and run locally and on CODE. (when running locally, run https://github.com/guardian/presence-indicator first)

compared to main, running `yarn audit --groups dependencies` will show a reduction.

fewer vulnerabilities than main

Should work as before.


dblatcher:
tested on CODE, API worked, UI loaded and was able to edit data.

davidfurey:
Looks reasonable to me. I tested creating a new piece of content in Workflow and editing the description.

davidfurey (approved):

PR #435, opened 2024-06-11
Add timeline and mini profiles creation option behind feature switches

Georges-GNM:
## What does this change?

Adds two feature switches to enable creating timeline and mini profiles articles from the create new dropdown. 

## How to test

Load the branch, preferably on code. Shift + F12 should show feature switches for timeline and miniprofiles, which should add the corresponding option to the create new dropdown. 

Trying to create a new timeline/miniprofiles should result in the article creation form showing the correct format. Actually hitting the create new article button and viewing the file in composer should bring up the intended template. 

## Images

image

image


public/components/feature-switches/feature-switches.js
```diff
@@ -40,8 +42,11 @@ class FeatureSwitches {
 }
 
 function wfFeatureSwitchesController ($scope, wfPreferencesService) {
-    const featureSwitchKeys = [];
-    $scope.readableNames = {}
+    const featureSwitchKeys = [timelineFeatureSwitchKey, miniProfilesFeatureSwitchKey];
+    $scope.readableNames = {
+        [timelineFeatureSwitchKey]: 'Timeline',
+        [miniProfilesFeatureSwitchKey]: 'Mini Profiles',
```
simonbyford, line 48: Sorry to nitpik but this should be `Mini profiles` (lower case p) throughout!

https://chat.google.com/room/AAAAw-4p-YQ/GgdcMpW-iUs/GgdcMpW-iUs?cls=10

public/lib/content-service.js
```diff
@@ -16,11 +16,17 @@ angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService
             class ContentService {
 
                 provideFormats(featureSwitches){
-                    const articleFormats = {
+                    let articleFormats = {
```
simonbyford, line 19: `const` only prevents reassignment using the assignment operator (i.e. =) so this isn't actually necessary!

simonbyford:
I see two weird ghost options when testing on CODE. Could this point to an issue with how feature switches are persisted?

Screenshot 2024-06-12 at 10 25 36

simonbyford:
With both feature switches disabled, I can still create Timeline and Mini profiles articles by creating a Q&A Explainer article then just selecting a different format in the dropdown. Niche case but worth protecting against?

Screenshot 2024-06-12 at 10 30 27

simonbyford (commented):
Looking good! I've left a few small comments

public/lib/content-service.js
Georges-GNM: Huh, that's right! Somehow haven't come across this particular operation before, but good to know 👍

public/components/feature-switches/feature-switches.js
Georges-GNM: Fair one, thought I'd gotten it all correct this time!

Georges-GNM:
Yep, think I've just pushed a fix 👍

public/components/feature-switches/feature-switches.js
```diff
@@ -54,10 +59,16 @@ function wfFeatureSwitchesController ($scope, wfPreferencesService) {
 
     const featureSwitches = new FeatureSwitches(getDefaultFeatureSwitchValues(), $scope.featureSwitchEntries)
 
-    const updateLocalFeatureSwitchValues = (featureSwitchResult) => {
-        const incomingFeatureSwitches = featureSwitchResult;
-        if (_.isObject(incomingFeatureSwitches)){
-            featureSwitches.update(incomingFeatureSwitches);
+    const updateLocalFeatureSwitchValues = (featureSwitchResult) => {   
+        const filteredIncomingFeatureSwitchResult = Object.keys(featureSwitchResult)
```
Georges-GNM, line 63: This should mean that regardless of the preferences stored in the database, the only feature switches that get used when updating the UI are the ones that match to a current feature switch as defined by the `featureSwitchKeys` array.

simonbyford: This seems to do the trick! 

Screenshot 2024-06-12 at 15 41 53

simonbyford (approved):
One little stray `console.log` but otherwise looking perfect, nice one!

public/components/stub-modal/stub-modal.js
```diff
@@ -137,6 +137,7 @@ function StubModalInstanceCtrl($rootScope, $scope, $modalInstance, $window, conf
     }
 
     $scope.articleFormats = articleFormatService.getArticleFormats();
+    console.log("articleFormats", $scope.articleFormats)
```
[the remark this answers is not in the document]
Georges-GNM: The log has been chopped 🪓

simonbyford (approved):

PR #436, opened 2024-06-25
Remove timeline feature switch

Georges-GNM:
We're ready to roll out the new format, which means removing the feature switch.

How to review and test:
Load the branch, you should be able to create a timeline article directly without having to activate a feature switch.

Mini profiles should still require one though.


simonbyford (approved):
Looks good to me! Have verified on CODE

PR #437, opened 2024-07-16
react to angular

dblatcher:
## What does this change?

This PR:
 - copies the react2angular setup from composer (flexible-content repo), 
 - uses it to define the page title block as a simple react component
 - starts a `style.ts` module to replicate the vars defined in the SCSS code

## How to test

Client app will still build and run as before, with no visible changes to the UI.

This PR does provide any direct improvements, but provides a pattern: 
 - for doing future frontend work in react
 - gradually phasing out angular on a future migration project 

 - does add extra moving parts to client side project with some old dependencies. Should ideally follow up with an upgrade to the older build tools.
 - there is some code duplication from composer (IE the react2Angular implementation)


public/react/TopToolbarTitle.tsx
```diff
@@ -0,0 +1,53 @@
...
+  display: "inline-block",
+  textAlign: "center",
+  lineHeight: "55px",
+};
+
+const titleStyle = css({
+  height: '100%',
+  width: dashboardSidebarWidth,
+  minWidth: dashboardSidebarWidth,
+  fontSize: 18,
```
KaliedaRik, line 19: My personal preference is to always give font sizes an explicit suffix - in this case `px`? - just in case

public/react/style.ts
```diff
@@ -0,0 +1,24 @@
+export const dashboardSidebarWidth = 160;
```
KaliedaRik, line 1: Similar to above comment - should this be '160px'?

KaliedaRik (dismissed):
Tested in local - I can see the top toolbar title component is landing with a react-derived class rather than an angular class.

Screenshot 2024-07-22 at 15 35 59

I've left a couple of (easily ignorable) comments in the code, for consideration. Other than that, happy to approve!

public/react/TopToolbarTitle.tsx
dblatcher: 👍 Makes sense - I managed to trip myself up earlier when I forgot that `px` is not the default unit for `lineHeight`, so all in favour of explicit units for typography.

public/react/style.ts
dblatcher: I'd tend to keep exported dimension as numbers rather than strings so they can be used in calculations easily. If the units need to be included for clarity, we could do that where the values are used eg  ```fontSize:`${dashboardSidebarWidth/4}px```

KaliedaRik (approved):

PR #444, opened 2024-08-20
Remove mini profiles feature switch

simonbyford:
We're ready to roll out the Mini profiles format, which means removing the feature switch.

### How to review and test

Run workflow locally or on CODE. You should be able to create a Mini profiles article without having to activate a feature switch.


rhystmills (approved):
Tested on CODE and working as expected 👍

PR #451, opened 2024-09-02
Remove workflow subscription and notification service

Fweddi:
## What does this change?

Recently we discovered the Workflow subscription service, introduced in https://github.com/guardian/workflow-frontend/pull/153 and https://github.com/guardian/editorial-tools-platform/pull/232, has been broken since the 20th Feb.

This is the feature where you can subscribe to pages in Workflow and receive notifications for them:

image

We could switch this feature back on, but it has a lot of difficult to resolve security vulnerabilities, and the feature is used sparingly (~14 editorial users were using Workflow notifications before 20th Feb). 
 
Instead, we have decided to remove this service. It's a good feature, and perhaps worth being revived in future, but it's not being used frequently enough to justify the maintenance costs.

This PR follows a [PR to tear down the infrastructure](https://github.com/guardian/editorial-tools-platform/pull/806).

### Note
Will need to remove "webpush.publicKey", "webpush.privateKey" from params store. 

Let's keep the dependencies for now and remove in a separate PR (#454).


dblatcher (approved):
Tested on CODE - works as expected, feature removed from the menu

PR #453, opened 2024-09-02
Mark Karma and JSDom as dev dependencies

Fweddi:
## What does this change?

We have some security vulnerabilities around Karma (and JSDom, AFAIK only used by Karma). Karma can't be upgraded beyond `6.4.4` as it is EOL.

Before we find a solution to move away from Karma, let's mark it properly as a dev dependency. This should remove some vulns in production. 

While we're here, fix the `braces` vuln in production by bumping some dependencies (it will remain in dev because of - you guessed it - Karma).

## How to test

Deploy to CODE, ???, profit.


rhystmills (approved):
Looks sensible - I wonder if we have any similarly mis-labelled dev dependencies in other repos and whether we could reorganise to get our production vulns down...

Deployed to CODE and all running as expected 👍

PR #454, opened 2024-09-03
Remove notification service dependencies

Fweddi:
## What does this change?
Removes dependencies used by the notification service - removed in #451. Should resolve a few difficult vulnerabilities! 

## How to test
Deploy to CODE and test the UI.


andrew-nowak (approved):
project code is gone, so removing the sbt config should be no-op, lgtm
(slightly surprised that sbt never complained that the project code was missing; I suppose because of some combination of: the project is `lazy`, wasn't `.aggregate`d and CI never called `compile`, `test` or other task on the project...)

PR #455, opened 2024-09-03
Upgrade angular to latest version of v1

Fweddi:
## What does this change?
Removes some vulnerabilities.

## How to test
Deploy to CODE and test UI (perhaps routing as well as we are updating the router library).


package.json
```diff
@@ -41,12 +41,12 @@
     "@emotion/react": "^11.11.4",
     "@types/react": "^17.0.76",
     "@types/react-dom": "^17.0.9",
-    "angular": "1.5.11",
-    "angular-animate": "1.5.11",
-    "angular-bootstrap-datetimepicker": "1.1.4",
```
Fweddi, line 46: We receive some warnings about using `angular` instead of `angularjs` (which is pinned to 1.X `angular`). No real gains to this change other than suppressing the warning (and avoiding accidental upgrades to more recent angular versions?)

rhystmills (dismissed):
All makes sense, deployed on CODE and working as expected 👍

rhystmills (approved):

PR #460, opened 2024-09-19
Upgrade Angular 1.5 to 1.6

Fweddi:
## What does this change?
Incremental upgrade towards Angular 1.8 to resolve high vulnerabilities.

## How to test

Deploy to CODE. Smoke test. Check issues spotted in #456 have been resolved.

We can rollback as before if needed. We should communicate to CP to make sure we hear of any issues as soon as we can.


public/components/editable-field/editable-field.js
```diff
@@ -165,9 +165,6 @@ function wfEditableTextFieldDirectiveFactory($timeout) {
                 $attrs.$addClass('editable__text-field--textarea');
             }
 
-            // resets / sets the ng-model-options (prevents default behaviour)
```
Fweddi, line 168: As of Angular 1.6, ng-model-options are  guaranteed to be defined (https://docs.angularjs.org/guide/migration#migrate1.5to1.6-ng-directives-ngModelOptions). Previously we relied on options being set as an empty object, which allowed us to unset the 'updateOnDefault' key. Now we need to pass this value manually.

Without this, updates are posted to the view and model concurrently. We want to post updates to the form view on input change, and to the form model on save.

Without this fix, inputs were being wiped on timeout / debounce. I'm still not quite sure why, but this seems to have worked.

dblatcher (approved):

PR #463, opened 2024-09-23
Upgrade Angular 1.5 to 1.6 [REDUX]

Fweddi:
## What does this change?

#460, with a fix for #462. 

We were applying the options field to all content fields; however this broke dropdown fields (presumably because in this case we _do_ want changes to update by default).

We are now more precise - we apply the options only on the editable field component (which is an input / text area).

## How to test
Smoke test on CODE - test dropdowns as well as inputs.


dblatcher (approved):
tested on CODE - the priority dropdowns work

PR #468, opened 2024-10-15
Log telemetry event when Workflow create button is clicked

Fweddi:
## What does this change?

Adds telemetry service to Workflow, which logs to: https://metrics.gutools.co.uk/d/ae0yg6hy27kzke/workflow-telemetry?orgId=1&var-Stage=CODE.

At a first pass this logs an event when a Composer article is created from the Workflow 'Create new' dropdown. 

In a further pass we can track other details we might want to log (e.g. commissioned length, missing commissioned length reason).

### Testing
This branch has been deployed to [CODE](https://workflow.code.dev-gutools.co.uk/dashboard). Create an article using the dropdown modal. You should see a network request to the user telemetry service. The event will be logged in https://metrics.gutools.co.uk/d/ae0yg6hy27kzke/workflow-telemetry?orgId=1&var-Stage=CODE.

### DIFF
--- a/public/lib/content-service.js
+++ b/public/lib/content-service.js
@@ -6,10 +6,11 @@ import './atom-workshop-service'
 import './http-session-service';
 import './user';
 import './visibility-service';
+import './telemetry-service';
 
-angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService', 'wfDateService', 'wfFiltersService', 'wfUser', 'wfComposerService', 'wfMediaAtomMakerService', 'wfAtomWorkshopService', 'wfPreferencesService'])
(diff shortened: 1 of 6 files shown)


public/lib/content-service.js
```diff
@@ -117,6 +118,8 @@ angular.module('wfContentService', ['wfHttpSessionService', 'wfVisibilityService
                                 updatedStub['status'] = statusOption;
                             }
 
+                            wfTelemetryService.sendTelemetryEvent("WORKFLOW_CREATE_IN_COMPOSER_TRIGGERED", { contentId: stub.id })
```
Fweddi, line 121: not sure if this stub ID is always set

dblatcher:
tried this on CODE - the network event was sent correctly and received an OK response, but I'm not seeing any data in the dashboard

Fweddi:
> tried this on CODE - the network event was sent correctly and received an OK response, but I'm not seeing any data in the dashboard

Whoops, forgot to save the dashboard settings. That should work now?

dblatcher (approved):

PR #474, opened 2024-10-23
Fix tag typing (stick to strings)

Fweddi:
## What does this change?

Telemetry events aren't being sent for this button. We are receiving errors in the console (in the network response):
```
{
    "status": "error",
    "message": "Incorrect event format",
    "data": [
        {
            "keyword": "type",
            "dataPath": "[1].tags['commissionedLength']",
            "schemaPath": "#/definitions/IUserTelemetryEvent/properties/tags/additionalProperties/type",
            "params": {
                "type": "string,number,boolean"
            },
            "message": "should be string,number,boolean"
        }
    ]
}
```

The commissionedLength is currently sometimes a number, sometimes null. Let's only add it as an event tag if it's defined, and make it a string to make sure that isn't the issue.

For the other tags, let's do the same - add them optionally if we sometimes expect them to be non-null.

Another version of this PR is filtering the tags for the non-empty values, if we want to do it that way. 

### Testing
I have deployed to CODE. The errors have resolved and the events are sending as expected.

### DIFF
--- a/public/lib/composer-service.js
+++ b/public/lib/composer-service.js
@@ -148,12 +148,12 @@ function wfComposerService($http, $q, config, $log, wfHttpSessionService, wfTele
 
         const tags = {
             contentType: getType(type),
-            displayHint: selectedDisplayHint,
-            commissionedLength,
             productionOffice: prodOffice,
-            commissioningDesk: commissioningDeskExternalName,
             priority: getPriorityName(priority),
         }
+        if(selectedDisplayHint) tags.displayHint = selectedDisplayHint;
+        if(commissionedLength) tags.commissionedLength = commissionedLength.toString();
+        if(commissioningDeskExternalName) tags.commissioningDesk = commissioningDeskExternalName;
         if(missingCommissionedLengthReason) tags.missingCommissionedLengthReason = missingCommissionedLengthReason;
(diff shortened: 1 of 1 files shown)


public/lib/composer-service.js
```diff
@@ -148,12 +148,12 @@ function wfComposerService($http, $q, config, $log, wfHttpSessionService, wfTele
 
         const tags = {
             contentType: getType(type),
-            displayHint: selectedDisplayHint,
-            commissionedLength,
             productionOffice: prodOffice,
-            commissioningDesk: commissioningDeskExternalName,
             priority: getPriorityName(priority),
```
Fweddi, line 152: These should always have a value.

Fweddi, line 152: contentType, productionOffice and priority should always have a value.

public/lib/composer-service.js
```diff
@@ -148,12 +148,12 @@ function wfComposerService($http, $q, config, $log, wfHttpSessionService, wfTele
...
         }
+        if(selectedDisplayHint) tags.displayHint = selectedDisplayHint;
```
davidfurey, line 154: Generally nicer to explicitly check for not null and not undefined rather than coercing to a boolean. Particularly relevant for commissionedLength, when you would want to send zero if it is zero.

Fweddi: I'm with you. Let me change that.

davidfurey (approved):
Looks good to me

PR #476, opened 2024-11-01
Surface missingCommissionedLengthReason field

Fweddi:
We recently made commissioned length mandatory in Workflow and gave a way for users to opt out by clicking the Breaking News button. 

This PR surfaces this choice in the Workflow UI (in the `Commissioned Length` column, renamed from the vague - if shorter - `Commission`).

### To test
Run locally or deploy to CODE alongside https://github.com/guardian/workflow/pull/1123 and https://github.com/guardian/flexible-content/pull/5070.

### DIFF
--- a/public/components/content-list-drawer/content-list-drawer.js
+++ b/public/components/content-list-drawer/content-list-drawer.js
@@ -332,23 +332,20 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
             };
 
             $scope.updateCommissionedLength = function (newValue) {
-              updateField("commissionedLength", newValue);
-              if (newValue === "") {
-                return wfComposerService.deleteField(
-                  $scope.contentItem.composerId,
-                  "commissionedLength"
-                );
-              } else if ($scope.contentItem.missingCommissionedLengthReason !== null) {
+              // Don't allow commissioned length to be unset
+              if(newValue === "") return;
+              if ($scope.contentItem.missingCommissionedLengthReason !== null && $scope.contentItem.missingCommissionedLengthReason !== undefined) {
+                  updateField("missingCommissionedLengthReason", null);
                   wfComposerService.deleteField(
                       $scope.contentItem.composerId,
                       "missingCommissionedLengthReason"
                   );
               }
-
-              return wfComposerService.updateField(
-                $scope.contentItem.composerId,
-                "commissionedLength",
-                newValue
+              updateField("commissionedLength", newValue);
+              wfComposerService.updateField(
+                  $scope.contentItem.composerId,
+                  "commissionedLength",
+                  newValue
               );
             };
 
--- a/public/components/content-list-drawer/content-list-drawer.js
+++ b/public/components/content-list-drawer/content-list-drawer.js
@@ -346,6 +346,14 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
               );
             };
 
+            $scope.formatReason = (missingCommissionedLengthReason) => {
+                const reasons = {
+                    "BreakingNews": "Breaking News",
+                }
+
+                return reasons[missingCommissionedLengthReason] || missingCommissionedLengthReason;
+            }
+
             /**
              * Revert deadline to previous state
              */
--- a/public/components/content-list-drawer/content-list-drawer.js
+++ b/public/components/content-list-drawer/content-list-drawer.js
@@ -338,7 +338,13 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
                   $scope.contentItem.composerId,
                   "commissionedLength"
                 );
+              } else if ($scope.contentItem.missingCommissionedLengthReason !== null) {
+                  wfComposerService.deleteField(
+                      $scope.contentItem.composerId,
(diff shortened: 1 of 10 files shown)


public/components/content-list-drawer/content-list-drawer.js
```diff
@@ -332,20 +332,31 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
...
             $scope.updateCommissionedLength = function (newValue) {
+              // Don't allow commissioned length to be unset
+              if(newValue === "") return;
+              if ($scope.contentItem.missingCommissionedLengthReason !== null && $scope.contentItem.missingCommissionedLengthReason !== undefined) {
+                  updateField("missingCommissionedLengthReason", null);
+                  wfComposerService.deleteField(
+                      $scope.contentItem.composerId,
+                      "missingCommissionedLengthReason"
+                  );
+              }
```
Fweddi, line 343: Happy for others to disagree with this choice. But there doesn't seem to be much benefit in allowing people to remove a commissioned length from a piece that has one set.

Theoretically it would allow users to correct a mistake, but in practice since we expect all (or almost all) pieces to have a commissioned length, it seems more likely that someone might mis-use this feature, i.e. by setting an arbitrary commissioned length on the Workflow form and then removing it to get round the mandate.

public/lib/column-defaults.js
```diff
@@ -311,6 +311,17 @@ const columnDefaults = [{
     isNew: true,
     isSortable: true,
     sortField: ['statusInPrint']
+},{
```
Fweddi, line 314: Re-ordered so the commissionedLength can be next to the webWordCount (and so more easily comparable). But keeping webWordCount and printWordCount next to each other.

This is the inverse order of the Ophan screen (wordCount / commissionedLength), but hard to balance all the needs.

public/components/content-list-drawer/content-list-drawer.js
Fweddi, line 343: Unset the missingCommissionedLengthReason when a commissioned length is set.

Fweddi, line 343: Happy for others to disagree with this choice. But there doesn't seem to be much benefit in allowing people to remove a commissioned length from a piece that has one set.

Theoretically it would allow users to correct a mistake, but in practice since we expect all (or almost all) pieces to have a commissioned length, it seems more likely that someone might mis-use this feature, i.e. by setting an arbitrary commissioned length on the Workflow form and then removing it to get round the mandate.

Fweddi, line 343: Unset the missingCommissionedLengthReason when a commissioned length is set.

public/components/content-list-drawer/content-list-drawer.js
```diff
@@ -332,20 +332,31 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
             };
 
             $scope.updateCommissionedLength = function (newValue) {
+              // Don't allow commissioned length to be unset
+              if(newValue === "") return;
```
Fweddi, line 336: Happy for others to disagree with this choice. But there doesn't seem to be much benefit in allowing people to remove a commissioned length from a piece that has one set.

Theoretically it would allow users to correct a mistake, but in practice since we expect all (or almost all) pieces to have a commissioned length, it seems more likely that someone might mis-use this feature, i.e. by setting an arbitrary commissioned length on the Workflow form and then removing it to get round the mandate.

public/lib/column-defaults.js
```diff
@@ -311,6 +311,17 @@ const columnDefaults = [{
...
+    name: 'commissionedLength',
+    prettyName: 'Commissioned Length',
+    labelHTML: 'Commissioned Length',
+    colspan: 1,
+    title: '',
+    templateUrl: templateRoot + 'commissionedLength.html',
+    template: commissionedLengthTemplate,
+    active: true,
+    isSortable: true,
+    sortField: ['missingCommissionedLengthReason', 'commissionedLength']
```
Fweddi, line 324: If you order the Commissioned Length field by ascending, you should see missingCommissionedLengthReason first.

Would be nice if we could order by descending to see very high Commissioned Length values, but this just shows empty values as highest. Perhaps some refactoring of the sort would help? (Should blank values be included or not in sort is the bigger question).

dblatcher (approved):

dblatcher:
> NB. The Commissioned Length editable field in Workflow (in the Furniture panel for a given stub) has a slightly horrible UX, at least on my browser. If you scroll from left to right while editing the field (e.g. trying to navigate Workflow at full width, with all columns), this sideway scrolling increases or decreases the Commissioned Length value. So if you're wondering why you wrote a value of 200, scroll to confirm, and then see it has a value of 208, you're not deranged...

I came across the same issue and spent a small amount of time looking for bugs in the api logic! 

It might be possible to fix this by capturing [wheel events](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event) with  the `<input>` in focus and preventing the them from bubbling to scroll with window, but I wouldn't  be too confident it would work nicely with angular.

PR #480, opened 2024-11-20
Update production office in Composer

Fweddi:
For a while now, changing the Production Office in Workflow would only update the stub in the Workflow DB.

This PR updates the Production Office field in Composer as well - using the settings endpoint (as Production Office is a setting, not a field).

### To test
In either local or CODE workflow, find the Management tab of an article and change the Production Office setting.

You should see two PUT network requests to the Production Office setting endpoint in Composer (one for preview, another for live).

The article in Composer CODE should update with the new value of Production Office. You can see this in three places:
(1) the Management tab
(2) the API debug endpoint - here you can see both draft and live settings
(3) the 'Track in Workflow' section of the sidebar (this reads from the Workflow DB)

Before this PR, only (3) would update, leading to confusion.

### DIFF
--- a/public/components/content-list-drawer/content-list-drawer.js
+++ b/public/components/content-list-drawer/content-list-drawer.js
@@ -320,6 +320,23 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
                 updateField("plannedNewspaperPageNumber", pageNumber, $scope.contentItem.plannedNewspaperPageNumber)
             }
 
+            $scope.updateOffice = (office) => {
+                // preview
+                wfComposerService.updateSetting(
+                    $scope.contentItem.composerId,
+                    "productionOffice",
+                    office,
+                )
+
+                // live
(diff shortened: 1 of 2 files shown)


public/components/content-list-drawer/content-list-drawer.js
```diff
@@ -320,6 +320,23 @@ export function wfContentListDrawer($rootScope, config, $timeout, $window, conte
...
 
+            $scope.updateOffice = (office) => {
+                // preview
+                wfComposerService.updateSetting(
+                    $scope.contentItem.composerId,
+                    "productionOffice",
+                    office,
+                )
+
+                // live
```
Fweddi, line 331: For fields / settings like Production Office, the draft / live distinction doesn't matter, so we update both (regardless of publication state).

simonbyford (approved):
Looks good and everything works as expected on CODE 🚢

PR #488, opened 2025-01-14
Fix sbt dependency graph workflow

emdash-ie:
## What does this change?

At the moment, this workflow is [failing on main](https://github.com/guardian/workflow-frontend/actions/runs/12430359668/job/34705650369) because it doesn’t explicitly install sbt, and github has updated the ubuntu-latest tag to the 24.04 runners which don’t include sbt by default.

To fix this, this commit updates the workflow by copying the version from https://github.com/guardian/fastly-cache-purger/pull/88, which PR was produced by a newer version of the gu-dependency-graph-integrator bot and includes an explicit installation of sbt (and updates to some action versions).

## How to test

To test this, we could modify the workflow to be triggered by this branch and confirm it runs successfully. Alternatively, we could just merge and see if it fixes it: since it's currently broken I don't think merging could make things any worse, and I'm confident this fixes the problem.

The dependency graph in github will be up to date, and the build on main will be marked as completely successful in the github UI.

### DIFF
--- a/.github/workflows/sbt-dependency-graph.yaml
+++ b/.github/workflows/sbt-dependency-graph.yaml
@@ -1,19 +1,28 @@
-name: Update Dependency Graph for SBT
+name: Update Dependency Graph for sbt
 on:
   push:
     branches:
       - main
-  workflow_dispatch: 
+  workflow_dispatch:
 jobs:
   dependency-graph:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


emdash-ie:
Happily the workflow run succeeded after merge: https://github.com/guardian/workflow-frontend/actions/runs/12766839912/job/35583974897

PR #489, opened 2025-01-16
Update panda to v7

emdash-ie:
## What does this change?

This PR updates the dependencies on [panda](https://github.com/guardian/pan-domain-authentication) to 7.0.0, following the instructions in https://github.com/guardian/pan-domain-authentication/issues/160.

I also took the opportunity to update the use of actions-riff-raff to v4 in https://github.com/guardian/workflow-frontend/pull/489/commits/9ef3deaa99a33c19948bd1de8424db76f6a30629.

## How to test

To test, I [deployed this PR to CODE](https://riffraff.gutools.co.uk/deployment/view/54ec3f48-738e-4acf-a3ac-94f12cf45959), and then confirmed that I can still visit https://workflow.code.dev-gutools.co.uk/dashboard. I also tried clearing my cookies and watching the network tab to make sure the oauth redirects happened and worked, which they did. Finally, I tried accessing the page from a private tab (where I wasn’t signed into my google account) and saw a prompt from google to log in.

Is there anything else I should check?

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -1,4 +1,4 @@
-import com.gu.pandomainauth.PanDomainAuthSettingsRefresher
+import com.gu.pandomainauth.{PanDomainAuthSettingsRefresher, S3BucketLoader}
 import com.gu.permissions.{PermissionsConfig, PermissionsProvider}
 import com.gu.workflow.api.{DesksAPI, SectionDeskMappingsAPI, SectionsAPI, StubAPI}
 import com.gu.workflow.lib.TagService
@@ -24,12 +24,10 @@ class AppComponents(context: Context)
 
   val config = new Config(context.initialConfiguration)
 
-  val panDomainRefresher = new PanDomainAuthSettingsRefresher(
+  val panDomainRefresher = PanDomainAuthSettingsRefresher(
     domain = config.domain,
     system = config.pandaSystem,
-    bucketName = config.pandaBucketName,
-    settingsFileKey = config.pandaSettingsFile,
-    s3Client = AWS.S3Client
+    S3BucketLoader.forAwsSdkV1(AWS.S3Client, "pan-domain-auth-settings")
   )
 
   val permissions: PermissionsProvider =
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,9 +10,12 @@ jobs:
     build:
         runs-on: ubuntu-20.04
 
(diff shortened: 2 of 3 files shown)


andrew-nowak (approved):
lgtm! thanks!

emdash-ie:
I've checked on workfow in PROD post-merge and I'm still able (and required) to log in. I also looked at the logs and didn't see anything very concerning: [a spike in WARN/ERROR logs right at deployment time](https://logs.gutools.co.uk/s/editorial-tools/app/r/s/JMHKu), but that has since gone away.

app/AppComponents.scala
```diff
@@ -24,12 +24,10 @@ class AppComponents(context: Context)
...
   val config = new Config(context.initialConfiguration)
 
-  val panDomainRefresher = new PanDomainAuthSettingsRefresher(
+  val panDomainRefresher = PanDomainAuthSettingsRefresher(
     domain = config.domain,
     system = config.pandaSystem,
-    bucketName = config.pandaBucketName,
-    settingsFileKey = config.pandaSettingsFile,
-    s3Client = AWS.S3Client
+    S3BucketLoader.forAwsSdkV1(AWS.S3Client, "pan-domain-auth-settings")
```
emdash-ie, line 30: I think I made a mistake here: the second parameter for `S3BucketLoader.forAwsSdkV1` is called `bucketName`, so presumably this should’ve been `config.pandaBucketName` rather than this literal string I copied from another example?

app/AppComponents.scala
```diff
@@ -24,12 +24,10 @@ class AppComponents(context: Context)
 
   val config = new Config(context.initialConfiguration)
 
-  val panDomainRefresher = new PanDomainAuthSettingsRefresher(
+  val panDomainRefresher = PanDomainAuthSettingsRefresher(
     domain = config.domain,
     system = config.pandaSystem,
-    bucketName = config.pandaBucketName,
-    settingsFileKey = config.pandaSettingsFile,
```
emdash-ie, line 31: Also, could this config option be removed if it’s no longer used?

app/AppComponents.scala
rtyley: Ah yes, that probably should have been `config.pandaSettingsFile` for the smallest possible change, tho' `"pan-domain-auth-settings"` is almost certainly the value `config.pandaSettingsFile` would give!

emdash-ie: I’ll prepare a new PR that uses the config value or removes it, and removes the settingsFileKey config value.

PR #491, opened 2025-01-22
model display hint, centralise special format definitions

dblatcher:
## What does this change?

 - Adds the "displayHint" property to the `Stub` model so that it can be posted to the backend when creating new stubs for articles with special formats - see https://github.com/guardian/workflow/pull/1130 for the backend & DB changes to support the new property.
- Creates a ts type `SpecialArticleFormat` and a central list of the current formats available. Uses this list to create a collection of helpers to replace the code that was using hard coded values for the names or values of the formats - this should allow us to more easily add new formats in future

### DisplayHint property
 - Run this branch and the workflow project on https://github.com/guardian/workflow/pull/1130 locally.
 - Create a new miniProfile article in workflow from the "create new" button
 - Create a new timeline article in workflow from the "create new" button
 - (create some normal articles if you don't already have some in your DB)
 - manual set the display hint filter by navigating to  https://workflow.local.dev-gutools.co.uk/dashboard?display-hint=miniProfiles
 - after it has loaded, the list of content should only contain your new miniProfile (note - if you have existing miniProfile pieces in your DB, these will not show as the displayHint was not populated when they were created)
 - navigate to https://workflow.local.dev-gutools.co.uk/dashboard?display-hint=timeline and the list loaded should only contain your new timline article

### special format list
 - Run this branch locally
 - comment out on of the formats in public/lib/model/special-formats.ts, or set its `behindFeatureSwitch` property 
 - the format should be removed from the UI everywhere the formats appear (eg "create new" drop down, stub form) but the others should work as before
 - add a new item to the list (an include its `value` in the `SpecialFormatContentType` type ), reload the page
 - your new format should appear in the UI alongside the real once

## Accessibility

### DIFF
--- a/public/lib/model/special-formats.ts
+++ b/public/lib/model/special-formats.ts
@@ -1,13 +1,13 @@
 import { ComposerContentType, ContentType, SpecialArticleFormat, Stub } from "./stub";
 
-const STANDARD_ARTICLE_FORMAT_LABEL = "Standard Article"; 
+const STANDARD_ARTICLE_FORMAT_LABEL = "Standard Article";
 
 const specialFormats: SpecialArticleFormat[] = [
-    { label: 'Key Takeaways', value: 'keyTakeaways', iconId: 'keyTakeaways' },
-    { label: 'Q&A Explainer', value: 'qAndA', iconId: 'qAndA' },
-    { label: 'Timeline', value: 'timeline', iconId: 'timeline' },
-    { label: 'Mini profiles', value: 'miniProfiles', iconId: 'miniProfiles' },
-    { label: 'Multi-byline', value: 'multiByline', iconId: 'multiByline', hidden: true },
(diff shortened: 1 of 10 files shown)


public/lib/model/special-formats.ts
```diff
@@ -0,0 +1,37 @@
+import { ComposerContentType, ContentType, SpecialArticleFormat, Stub } from "./stub";
+
+const specialFormats: SpecialArticleFormat[] = [
```
waisingyiu, line 3: It's a great idea, centralising the list of special formats in this constant.

waisingyiu (approved):
Thank you David for your great work on adding display hint to the model and refactoring code around special format!  Happy to approve.

Just want to check if I got it right.  I guess the next step is to backfill the display hint column of the workflow database before we can add the special format to the content type filter?

PR #494, opened 2025-01-30
Fix minor panda update mistakes

emdash-ie:
## What does this change?

When recently updating panda to v7 in commit ce587649252e83a1ff2b53003e890987c0f238f2 (PR https://github.com/guardian/workflow-frontend/pull/489), I followed the example in https://github.com/guardian/tagmanager/pull/535/files#r1767036487 slightly too closely, replacing the use of the config key with a hard-coded string by mistake.

This PR restores the use of the config key, and removes a config key that that commit removed the only use of.


waisingyiu (approved):
Thank you Emily, happy to approve.

PR #495, opened 2025-01-30
display hint put endpoint

dblatcher:
## What does this change?

Adds a new PUT endpoint for the recently added displayHint property.

adds the value of the display hint as a data attribute on the table cell displaying the content-type icon. This is to verify that thee value is set for debugging only. We will not be read to use the property in a user-visible way (EG by using an icon matching the article special formats, or indicating a piece flagged as "immersive") until it is being populated for all new Stubs the existing Stubs have been backfilled to include the display hints of the composer piece.

corresponding change in Workflow : https://github.com/guardian/workflow/pull/1132 

## How to test

This PR won't have an effect until an application can call the new end point. We can defer merging this PR until a composer Branch is available to test it with.

It will be possible to PUT changes to the display hint from Composer, using the control in the "advanced" tab - once Composer has been updated to make the PUT request to workflow as a side effect of changing it in composer

Should be safe - the end point follows an existing pattern.

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -165,6 +165,18 @@ class Api(
     )}
   }
 
+  def putStubDisplayHint(stubId: Long) = {
+    def getDisplayHintOpt(input: String): Option[String] = if(input.length > 0) Some(input) else None
+    APIAuthAction.async { request =>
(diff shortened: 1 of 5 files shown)


waisingyiu (approved):
Great work!

PR #502, opened 2025-02-26
filter by supported display hints

dblatcher:
## What does this change?

updates https://github.com/guardian/workflow-frontend/pull/500 to also include icons and filtering by 'immersive', 'photoEssay' and 'numberedList' using the icons from @paperboyo .

Also renames the constants and functions for the listElement-based formats to be called "list element formats" rather than "special formats" since this term is more consistent with the naming pattern in Composer and distinguishes them more clearly from the  other display hints.

## How to test

testing notes on 
updates https://github.com/guardian/workflow-frontend/pull/500 still apply, but should also include some articles and interactives with the display hint set using the panel in Composer's "Advanced" tab.

Can filter by all display hints. Would be relatively easy to add new items to the list

## Images

https://github.com/user-attachments/assets/e6b33374-4526-4bbc-aa4b-58ebdc642a3c


emdash-ie:
Testing locally, the three new filters (immersive, photo essay, numbered list) on the left don't seem to work: nothing is displayed as matching the filter even when I have created three articles and set their display hint (in the advanced tab in composer). Is there something else I need to do to get pieces to be recognised as these types, or does this does suggest there's something wrong? (Will start looking into the code to see if I can figure out why it's not working, but thought I should flag it early in case you're able to tell more quickly than me!)

dblatcher:
> Testing locally, the three new filters (immersive, photo essay, numbered list) on the left don't seem to work: nothing is displayed as matching the filter even when I have created three articles and set their display hint (in the advanced tab in composer). Is there something else I need to do to get pieces to be recognised as these types, or does this does suggest there's something wrong? (Will start looking into the code to see if I can figure out why it's not working, but thought I should flag it early in case you're able to tell more quickly than me!)
> 
> (When running locally, I have workflow-frontend on this branch, workflow on the branch from [guardian/workflow#1134](https://github.com/guardian/workflow/pull/1134), and composer on main)

Sorry, I should have included extra notes about testing locally:
 - https://github.com/guardian/workflow/pull/1134 needs to be running locally
 - Local Composer will only send updates to workflow if the feature switch (shift-f12 for the menu) "requests to local services" is enabled - if you have this enabled, you should see network requests going to workflow.local when you change the display hint, if not on Composer only updates its own backend (this is so you can run Composer locally without also having workflow running, which is usually a good thing)

If you've already got that in place, do give me a shout if you want to pair to track down the issue.

emdash-ie:
> Sorry, I should have included extra notes about testing locally:

Ah, thank you!

>     * Local Composer will only send updates to workflow if the feature switch (shift-f12 for the menu) "requests to local services" is enabled - if you have this enabled, you should see network requests going to workflow.local when you change the display hint, if not on Composer only updates its own backend (this is so you can run Composer locally without also having workflow running, which is usually a good thing)

This is the bit I was missing, I'll test with that feature switch enabled now.

PR #506, opened 2025-03-06
Add type guard for value expected to be a string

emdash-ie:
## What does this change?

[A recent error in Sentry](https://the-guardian.sentry.io/issues/6362818459/?alert_rule_id=56564&alert_timestamp=1741252090194&alert_type=email&notification_uuid=58661c31-1ebe-47a1-927e-d679b91117f9&project=33134&referrer=alert_email) appears to indicate that sectionsString here is truthy but not a string. Since the code seems to expect it to be a string, this PR adds a type guard to make sure we only use the value if it is a string.

It seems this value comes from the request URL, though I’m not 100% sure how. I’m also not sure how it came to be a non-string value.

### DIFF
--- a/public/layouts/dashboard/dashboard-toolbar.js
+++ b/public/layouts/dashboard/dashboard-toolbar.js
@@ -29,7 +29,10 @@ angular.module('wfDashboardToolbar', ['wfFiltersService', 'wfDateService', 'wfPr
 
         function buildSelectedSections () {
             var sectionsString = wfFiltersService.get('section');
-            var sectionsStringArray = sectionsString ? sectionsString.split(',') : [];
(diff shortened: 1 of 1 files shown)


emdash-ie:
Maybe the non-string section value [came from preferences](https://github.com/guardian/workflow-frontend/blob/efca1b46a01db701bb675139430219fed8f079a7/public/layouts/dashboard/dashboard-toolbar.js#L45)? I’m not sure how the preferences work, and haven’t really been able to test it locally.

liamness (approved):

emdash-ie:
I’ll mark the problem as resolved in Sentry as well, see if it comes up again

PR #510, opened 2025-04-21
Add the tools audit tracking pixel to Workflow

SHession:
## What does this change?

Adds the [User Telemetry tracking pixel ](https://github.com/guardian/editorial-tools-user-telemetry-service?tab=readme-ov-file#user-telemetry-client), to audit the use of the Workflow frontend.  Results will be visible in the [tools audit dashboard](https://metrics.gutools.co.uk/d/aegzv59r9eghsd/tool-audit?orgId=1&var-domain=workflow.gutools.co.uk&var-path=All&var-interval=1d&var-chosen_apps=All&from=now-30d&to=now).

## How to test

Testable locally and in CODE. Navigating to different pages should result in outgoing image requests to user-telemetry URLs.

We are able to audit the use of the Workflow tool.

### DIFF
--- a/app/views/layout.scala.html
+++ b/app/views/layout.scala.html
@@ -32,6 +32,17 @@
 
 @content
 
+<!-- Tracking pixel for telemetry -->
+<script>
+  const domainMatch = /^.*\.(?<environment>local|code)\.dev-gutools\.co\.uk$|^.*\.gutools\.co\.uk$/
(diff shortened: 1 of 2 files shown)


waisingyiu (approved):
Nice work, thanks Sam!

I've validated the changes locally and I could see the HTTP calls to the telemetry endpoint with the correct query params.

PR #516, opened 2025-05-26
Update AngularJS 1.5.11 to 1.8.x

SHession:
## What does this change?

This PR attempts to address some known vulnerabilities with older version of AngularJS. 

| Before ([CODE](https://workflow.code.dev-gutools.co.uk/dashboard?status=Writers,Desk,Production%20Editor&incopy=false)) | After ([local](https://workflow.local.dev-gutools.co.uk/dashboard?desk=63&status=Writers,Desk,Production%20Editor&incopy=false)) |
|--------|--------|
| image | image |
| image | image |

## How to test

Testable locally, we should also deploy to CODE to check for any issues. Several features depend on integrations which may not work locally without additional setup. 

## How can we measure success?

Fewer High and Medium severity vulnerabilities.

PR #519, opened 2025-06-12
Catch errors with support logger requests to prevent error feedback loop

SHession:
## What does this change?

This PR attempts to fix reported issues of Workflow freezing. I have been able to identify a possible cause of these issues by blocking requests to the `/support/logger` endpoint using the browser network controls. If an error then occurs, a feedback loop will take place which will cause the application to crash.

It is suspected this issue was a result of #516, most likely this https://docs.angularjs.org/guide/migration#-http- 

and

 > Reporting possibly unhandled promise rejections that would otherwise go unnoticed.

https://docs.angularjs.org/guide/migration#migrating-from-1-5-to-1-6

The docs therefore suggest these aren't fundetmentally new errors, but create a new failure mode by being reported more accurately. 

This PR attempts to resolve the issue by catching the errors and logging a warning. This should prevent the feedback loop but will also prevent the error from being logged.

## How to test

Testable locally or on CODE. Block the /support/logger endpoint and observe when errors take place a feedback loop occurs. On this branch the issue should result in warnings logs in the browser console.

Fewer instances of Workflow freezing! 

Angular JS in general...

### DIFF
--- a/public/lib/logger.js
+++ b/public/lib/logger.js
@@ -21,10 +21,13 @@ logger.factory('logger', ['$injector', function($injector) {
             fields: fields
         };
 
-        return $http({method: 'POST',
-               url: LOG_URL,
-               data: data
-              });
+        return $http({
+            method: 'POST',
+            url: LOG_URL,
+            data: data
+        }).catch(function onError() {
+            console.warn("/support/logger request failed")
+        });
     }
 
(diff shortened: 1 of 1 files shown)


dblatcher (approved):
LGTM - great catch

PR #528, opened 2025-08-08
Update scala version to 2.13.16

emdash-ie:
This app is currently using Scala 2.13.12, which [is now nearly two years old](https://www.scala-lang.org/download/2.13.12.html). This commit updates the scala version to 2.13.16.

(I noticed this and decided to change it because my editor integration metals recently started complaining about this version.)

I think we should be ok to rely on CI plus a successful CODE deploy for testing this PR: I would expect scala version changes to cause compilation errors or total failure of the app, not more subtle bugs.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,7 +5,7 @@ import scala.sys.env
 
 Test / parallelExecution := false
 
-val scalaVersionNumber = "2.13.12"
+val scalaVersionNumber = "2.13.16"
 
 val buildInfo = Seq(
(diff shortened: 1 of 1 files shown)


emdash-ie:
> The only thing I can see is that we should upgrade sbt to at least 1.9.9 to fix the `console` task on these scala versions: I’ll do that separately.

Have raised https://github.com/guardian/workflow-frontend/pull/529

PR #532, opened 2025-10-07
Update @guardian/user-telemetry-client to 1.2.1

liamness:
## What does this change?

I don't think there are any actual changes in behaviour since v1.1.0, and the adjustment to use a lockfile in the upstream project won't add any appreciable extra security here as we were already using one in this project, just seems like a good idea to keep the dependency up to date.

## How to test

Not sure. There are very few usages of `sendTelemetryEvent()` in this project it seems. I will figure out the steps required to send `WORKFLOW_COMMISSIONED_LENGTH_SUGGESTION_PRESSED` and then put reproduction steps here.

User telemetry continuing to work

## Visual testing

Changes to the [Flexible Frontend](../flexible-frontend/README.md) project should be tested visually using Chromatic. Add the `run_chromatic` label to pull requests to trigger a visual test run.

### Visual Testing Required?

- [ ] Yes
- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   },
   "dependencies": {
     "@emotion/react": "^11.11.4",
-    "@guardian/user-telemetry-client": "^1.1.0",
+    "@guardian/user-telemetry-client": "1.2.1",
     "@types/react": "^17.0.76",
     "@types/react-dom": "^17.0.9",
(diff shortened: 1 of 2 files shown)


SHession (approved):
Tested on CODE, LGTM!

PR #566, opened 2026-03-25
Address vulnerable npm dependencies

SHession:
## What does this change?

Address vulnerable dependencies currently included in the project (although we expect these to be dev dependencies):
https://github.com/guardian/workflow-frontend/security/dependabot/224
https://github.com/guardian/workflow-frontend/security/dependabot/230

Tested locally and passes CI. I will test on CODE when it is available.

Project vulnerabilities addressed!

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "karma-sinon": "^1.0.5",
     "karma-webpack": "^5.0.1",
     "mini-css-extract-plugin": "^2.9.1",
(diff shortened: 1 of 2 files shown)


emdash-ie (approved):
Looks good! Can see the two vulnerable dependencies are removed from the lockfile, and happy it’ll work since you’ve tested in CODE.

PR #569, opened 2026-04-13
Fix setting the book section and publication date

dblatcher:
## What does this change?

Fixes the methods in "contentListDrawerController" for setting the book publication date and selecting a book section tag from the "management" tab of the content drawer.

In both cases, the bug was that controller method was trying to access the property from it's own $scope, but the values are actually held on $scope.$parent for the controls that have them methods attached.

running locally, you can now open the draw for a piece of content without a book section and:
 - type into the section input and get the autocomplete suggestions from the tag search, choose a tag to confirm
 - into a date and successfully submit the value

when on main:
 - the input will not respond as the method never fetches a list of tags (since the query string is not being accessed)
 - submitting a date will do nothing in the UI and show an error in the console (undefined value used instead of the date input.) 

Fixes  broken interactions

Should be safe - was previously broken

## Images

Screenshot 2026-04-13 at 15 34 20

## Accessibility

### DIFF
--- a/public/components/content-list-drawer/content-list-drawer.html
+++ b/public/components/content-list-drawer/content-list-drawer.html
@@ -414,7 +414,7 @@
                                 </button>
                                 <div class="editable--edit editable--inline autocomplete" ng-show="showBookTagPicker">
                                     <input class="text-input fullwidth"
-                                        ng-model="$parent.bookSectionQuery"
+                                        ng-model="bookSectionQuery"
                                         placeholder="Enter a book section…"
                                         name="bookSection"
                                         autocomplete="off"
@@ -463,7 +463,7 @@
                                         <div
                                         wf-date-time-picker
                                         wf-update-on="enter"
-                                        ng-model="$parent.currentPublicationDatePickerValue"
+                                        ng-model="currentPublicationDatePickerValue"
                                         wf-on-cancel="publicationDateTimeEdit = false; revertPlannedPublicationDate()"
                                         wf-cancel-on="blur"
                                         wf-on-submit="updatePlannedPublicationDate(); publicationDateTimeEdit = false;"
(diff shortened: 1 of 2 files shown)


liamness:
Looks like there's some other places that use these fields as though they're in the current scope. See the method `revertPlannedPublicationDate`, and also some cleanup logic for these fields, when we show / hide a row:

https://github.com/guardian/workflow-frontend/blob/1c6fcff6220f4763a365a22655e1817f1fb8775a/public/components/content-list-drawer/content-list-drawer.js#L95-L108

Could moving these fields down from the parent scope also solve this? I don't see them being accessed in any other scopes, so it seems unnecessary, perhaps I'm missing something though.

Have tested that this fixes setting the publication date, haven't checked that it fixes setting the book section as that appears to require being able to run tag manager (or pointing it at the CODE instance, perhaps). I'll have another look at that later.

dblatcher:
> Looks like there's some other places that use these fields as though they're in the current scope. See the method `revertPlannedPublicationDate`, and also some cleanup logic for these fields, when we show / hide a row:
> 
> https://github.com/guardian/workflow-frontend/blob/1c6fcff6220f4763a365a22655e1817f1fb8775a/public/components/content-list-drawer/content-list-drawer.js#L95-L108
> 
> Could moving these fields down from the parent scope also solve this? I don't see them being accessed in any other scopes, so it seems unnecessary, perhaps I'm missing something though.
> 
> Have tested that this fixes setting the publication date, haven't checked that it fixes setting the book section as that appears to require being able to run tag manager (or pointing it at the CODE instance, perhaps). I'll have another look at that later.

👍  In retrospect, I just assumed that it was the controller that was "wrong" rather than the template for some reason. Now you mention it, I don't see a reason why the template should be modelling `$parent.currentPublicationDatePickerValue` and `$parent.bookSectionQuery` with ng-model rather than using the value on the controller's own scope.

I just tried again with the template  changed instead of the method and that works too - also the intended behaviour of clearing the input values after submitting or cancelling works now.

liamness (approved):
Tested again locally and all seems well

PR #574, opened 2026-05-08
intended audience column

dblatcher:
## What does this change?

Part of https://github.com/guardian/workflow/issues/1452

Creates the API routes for receiving "intendedAudience" updates from composer and passing them to the backend to persist the change.

Adds a feature switch to enable a new "intended audience" column

Installs the Stand component library and uses the react2Angular script to create a wrapper component that derives the values needs to display the `IntendededAudienceSignifier` component.

In determining the value for the "source" (where the article was written, UK, AUS or US), the wrapper component will attend to parse the names of the Commissioning Desk to determine  the appropriate value - if it cannot, it will fallback to using the Production Office.

Running locally with these branches:
https://github.com/guardian/workflow/pull/1466
https://github.com/guardian/flexible-content/pull/6517
https://github.com/guardian/workflow-frontend/pull/574

On an article tracked in workflow, changing the intended audience in Composer will update the column in workflow (turn on the feature switch in Composer to reveal the UI)

Disabling the feature switch (shift-f12) will remove the column from the view and the dropdown at the end of menu to toggle columns on or off.

Enables visualisation of the audience in workflow. The feature switch allows us to add more features without users seeing the new UI in production until is it rolled out.

The logic for deriving the "source" from the available data is imprecise - this should be sufficient for now, but we may need to change how the commissioning desk tags are modelled in Workflow to provide better data - at present, the application is only aware of the tag ID and internal name.

## Images

https://github.com/user-attachments/assets/f18df202-60d0-475a-a3ae-562158725576


public/components/feature-switches/feature-switches.html
```diff
@@ -4,7 +4,7 @@
   </tr>
   <tr ng-repeat="switch in featureSwitchEntries">
     <td>{{readableNames[switch[0]]}}</td>
-    <td><input type="checkbox" ng-model="switch[1]" ng-click="updateFeatureSwitchPreference(switch[0], switch[1])"></td>
+    <td><input type="checkbox" ng-model="switch[1]" ng-click="updateFeatureSwitchPreference(switch[0], !switch[1])"></td>
```
liamness, line 7: So this would just set the same value instead of actually toggling the switch on / off?

public/app.js
liamness, line 221: Have you found that this triggers your devtools to open? I think we have an `event.preventDefault()` in Composer to stop that from happening.

liamness (commented):
Have only looked at the code, all seems to make sense, not tested yet

Also started a review and then forgot to submit so these comments would be visible...

public/components/feature-switches/feature-switches.js
```diff
@@ -41,9 +41,11 @@ class FeatureSwitches {
 function wfFeatureSwitchesController ($scope, wfPreferencesService) {
     const featureSwitchKeys = [
         // e.g. 'multiByline'
+        'intendedAudienceColumn'
     ];
     $scope.readableNames = {
         // e.g. 'multiByline': 'Multi-byline',
+        'intendedAudienceColumn': 'show itended audience'
```
liamness:
```suggestion
        'intendedAudienceColumn': 'Show Intended Audience'
```

public/components/feature-switches/feature-switches.html
dblatcher: Yes - I was a bit confused by that, but we haven't had any available feature switches since Feb 2025 (https://github.com/guardian/workflow-frontend/commit/e49824222f38428a5c70d2d23020325aeb9e0458) so it probably hadn't come up.

I think between then and now there was a PR to change the version of Angular - that might have had an impact on how "ng-model" works maybe.

public/app.js
dblatcher: I hadn't noticed that, but yes, Shift-F12 also the devTools shortcut in firefox - in composer, the `KeyboardShortcutService` uses `preventDefault` for all keyDowns that have a matching shortcut.

Will add the event.preventDefault() here too.

liamness (approved):

liamness:
One thing that I think _is_ worth considering, although not in this PR, is that the linting setup in this project is in a bit of a messy state. Running `yarn run lint` locally produces errors, but this check doesn't run on the CI. I can also see errors in my IDE (in `IntendedAudienceWrapper.tsx`) that aren't picked up by `yarn run lint` for some reason.

PR #575, opened 2026-05-18
create stubs with intended audience

dblatcher:
## What does this change?

Adds a drop-down to the "create article" form for selected an "Intended audience" if the the "intended audience" feature switch is on. The options for which are:
- Don't know
- Domestic for Global
- Domestic for Domestic
- Global for Global

If the feature switch is on, not selected an option will  give a validation warning (preventing users from submitting the form)

When the form is submitted, the selected option and the `stub.productionOffice`  is used to populated the `stub.intendedAudience` string value, which represents a list of the tracking/audience tags for the article as below:
- Don't know:  (no audience tags)
- Domestic for Global: **tracking/audience/global** and the tag corresponding to the production office eg **tracking/audience/au**
- Domestic for Domestic : the tag corresponding to the production office eg **tracking/audience/au**
- Global for Global: just **tracking/audience/global** 

The list of tags is serialised in `stub.intendedAudience` as a comma separated list of slugs - eg **tracking/audience/global** and  **tracking/audience/us** would be "global,us".

When POSTing the create request to Composer, the numerical ID's of the audience tags are included in the "tracking" path param, along with the ID for the commissioning desk tags.

This is a change to how `stub.intendedAudience` was set in https://github.com/guardian/flexible-content/pull/6517 and https://github.com/guardian/workflow-frontend/pull/577, so the way that it is used to derive the props for the `IntendedAudienceSignifier` component in the stub table has been updated too.

If testing locally, adjust app/config/Config.scala to point to Tagmanager CODE:
```
lazy val tagManagerUrl: String = "https://tagmanager.code.dev-gutools.co.uk"
```
The feature relies on the correct list of tracking/audience tags existing so the application can look-up the numerical ID's for the create request to Composer.

With the feature switch on, should be able, the "create" form should require you to select an option for "intended audience" ("Don't know" is accepted).
When you create the article: 
 - it should be added to Composer, with the expected tracking/audience tags, as well as the commissioning desk tag, if any  - see https://composer.{stage}.dev-gutools.co.uk/api/content/{contentId}/preview/taxonomy/tracking
 - the `IntendedAudienceSignifier` should render correctly in the tag table.

With the feature switch off, the form should look and behave as it does on main.

Can choose the intended audience when creating articles from workflow, the composer article will have the corresponding audience tags

Should be no change to users unless the feature switch is on, but should test creating articles with and without to be sure.

We have another issue open to make the corresponding modelling changes in Composer.

## Images

Screenshot 2026-05-20 at 16 17 54


public/components/stub-modal/stub-modal.html
```diff
@@ -42,13 +42,41 @@ <h4 class="modal-title">{{ modalTitle }}</h4>
...
+                            <p>Selecting commissioning info is temporarily unavailable</p>
+                        </div>
+                    </div>
+                    <div ng-if="isIntendedAudienceEnabled">
+                        <div class="label-and-icon">
+                            <label for="stub_intended_audience">
+                                <span>Intended audience</span>
+                                <span>*</span>
+                            </label>
+                            <i class="content-list-head__heading-icon--tooltip" wf-icon="tooltip" title="TBC"></i>
```
emdash-ie, line 62: Is this ok to still be TBC? Since it’s behind the feature switch I think it’s ok to merge as-is, but I wanted to flag it because I don’t know who we’re waiting on

emdash-ie (approved):
Looks good to me! Tested locally and in CODE and I can see it works. The tracking tags aren’t displayed anywhere in the composer UI at the moment though, are they? (I couldn’t see them even after enabling the feature switch, but I may not have been looking in the right place.)

public/components/stub-modal/stub-modal.html
dblatcher: Will check in on that - we have another task to update the version of Stand to get the new version of the signifier, so hopefully will add the text as part of the same PR.

dblatcher:
> Looks good to me! Tested locally and in CODE and I can see it works. The tracking tags aren’t displayed anywhere in the composer UI at the moment though, are they? (I couldn’t see them even after enabling the feature switch, but I may not have been looking in the right place.)

The tracking/commissioning-desk tags are show under Furniture, but no, the new tracking/audience tags aren't reflected in the Composer UI as yet - that's an upcoming change.

PR #577, opened 2026-05-19
intended audience  fixed

dblatcher:
## What does this change?

Reinstates https://github.com/guardian/workflow-frontend/pull/576 and fixes the bug came up in production - that user's column's preferences were not loading and instead being set to the default (all enabled)

The bug came from the use of `wfPreferencesService.getPreference('featureSwitches')` in "wfColumnService" to check if the new feature switch is enabled - this method returns a promise that will `reject` if the preferences data for the user from the preferences API  does not have a 'featureSwitches' key, which will be the case for any user who has never toggled a feature switch in workflow frontend (IE for most non-developers).

To fix that issue, this PR adds an extra method, `wfPreferencesService.getAllPreferences`, which returns a promise that that returns the full set of preferences data, so will not reject for missing keys. 

"wfColumnService" now uses `wfPreferencesService.getAllPreferences` to get both the user's feature switch preferences and column configuration (also a preference) from one request, instead to two nested promises from `getPreference`

Ran locally to check that the preferences for which columns to display are successfully loaded - to set the preferences,  
To change the column preferences, scroll the the right edge of the table and click the columns button, toggle off some colums with the checkboxes and click the reload button at the bottom of the list:
Screenshot 2026-05-20 at 09 20 18

Ideally, this PR should be tested by someone for whom the CODE preferences DB has no featureSwitch data for the workflow app - IE has no data at `https://preferences.code.dev-gutools.co.uk/preferences/FIRSTNAME.LASTNAME@guardian.co.uk/workflow/featureSwitches` - to replicate a typical user profile.

There isn't a easy way to unset/delete this data with the preferences API once it is set.

Also (to verify the feature in https://github.com/guardian/workflow-frontend/pull/576 still works)
 - with the feature switch disabled, the "intended audience" column should not be displayed, or selectable in the columns drop down
 - with the feature switch enabled, the "intended audience" column should be in the columns drop down and displayed when ticked
 - the signifier should display in the column if the article was tracked in workflow from composer with the Composer intended audience feature switch on.

Can put out the change without the bug from https://github.com/guardian/workflow-frontend/pull/576

The featureSwitch logic is still somewhat obscure and coupled to an Angular directive - as a follow up, it would be worth extracting the logic to a service class in typescript and making it available as an Angular service module.


public/lib/preferences-service.js
```diff
@@ -139,6 +145,32 @@ angular.module('wfPreferencesService', [])
                     }
                 }
 
+                 /**
+                 * Gets all preferences. 
+                 * 
+                 * If preferences have already been fetched, resolve a promise with the preference
+                 * requested, else return the promise of the request to the preferences app that will resolve
+                 * with preferencea
```
liamness:
```suggestion
                 * with preferences
```

liamness (dismissed):
Tested on CODE by deleting the features switches record for my user in the DynamoDB table—could still change which columns are visible.

1 minor spelling-related comment

liamness (approved):
Haven't retested, but changes since last review look minor and sensible

PR #582, opened 2026-06-01
update Stand, replace IntendedAudienceSignifier prop logic

### LINKED ISSUE #579: change to new version of Stand Signifier component
We have a new version of the intended audience signifier component with a different set of props.
Should update workflow-frontend to use the new version (will entail changing the log for converting the tags to signifier props)

We also still have "tbc" for the tool-tip titles. Have asked design for the text to use.

ALSO - do not render the "don't knows" in the tag table!

dblatcher:
## What does this change?

Updates the version of Stand used in workflow frontend and uses the `mapTagsToSourceAndTarget` function for deriving the props for the `IntendeAudienceSignifier`.

The new logic now parses the "intendedAudience" string value of the stub.externalData, which represents a list of audience tag slugs (eg "uk", "global", "us,global") to the `source` and `target` props which determines the icons displayed, eg :

| stub.externalData.intendedAudience | source | target|
|-----------|---|---|
"uk".           | UK|UK
"global".     |global|global 
"us,global" |US|global|

The stub table will no longer display the "Don't know" signifier if the audience is not defined.

Ran locally with the feature switch on and created some new stubs to check the display logic

Keeps the implementations of the component in step.

When we transition to using the array of tracking tags instead of the `stub.externalData.intendedAudience` string , it should be relatively easy to adapt the logic for deriving the props.

Just a display change, should be safe.

### DIFF
--- a/public/components/content-list-item/templates/intended-audience.html
+++ b/public/components/content-list-item/templates/intended-audience.html
@@ -2,8 +2,7 @@
   class="content-list-item__field--intended-audience"
   title="Intended Audience: {{ contentItem.intendedAudience || 'none' }}"
 >
-  <intended-audience-signifier 
-   production-office="contentItem.office"
-   intended-audience="contentItem.intendedAudience"
+  <intended-audience-signifier
+    intended-audience="contentItem.intendedAudience"
   ></intended-audience-signifier>
 </td>
--- a/public/components/content-list-item/templates/intended-audience.html
+++ b/public/components/content-list-item/templates/intended-audience.html
@@ -1,6 +1,5 @@
 <td
   class="content-list-item__field--intended-audience"
-  title="Intended Audience: {{ contentItem.intendedAudience || 'none' }}"
 >
   <intended-audience-signifier
     intended-audience="contentItem.intendedAudience"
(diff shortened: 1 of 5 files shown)


public/components/content-list-item/templates/intended-audience.html
```diff
@@ -2,8 +2,7 @@
   class="content-list-item__field--intended-audience"
   title="Intended Audience: {{ contentItem.intendedAudience || 'none' }}"
```
liamness: I think this line ought to go? As it will display the raw value which shouldn't be exposed to users

liamness (commented):
Tested locally, looks good, just one thing that I think needs changing.

The codeql failure is odd, seems like you can't even retry it? Maybe push a dummy commit and see if it succeeds then.

dblatcher:
> Tested locally, looks good, just one thing that I think needs changing.
> 
> The codeql failure is odd, seems like you can't even retry it? Maybe push a dummy commit and see if it succeeds then.

Thanks - have made the suggested change and the codeql worked this time - looking at the error, it sounded like a network issue for the box running the checks. I opened and closed a test PR yesterday to make sure it wasn't anything to do with the branch and got the same issue

Not being able to re-run it was a bit of a concern - could be quite a blocker if we needed to get a quick fix out. From the [github doc](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/re-run-workflows-and-jobs) page, I'm not sure why we can't retry it. 🤔

liamness (approved):

PR #588, opened 2026-06-17
Fix casing issue with intended audience signifier

liamness:
## What does this change?

Adds logic to transform the casing of the `intendedAudience` field to lower case before attempting to create "tags" which are passed into `mapTagsToSourceAndTarget()`. I don't quite understand why it's upper case because the PUT requests setting this field from Composer sends a lower case value. But I suppose we can figure this out later (and we eventually want to move away from using this field).

It has not! I currently cannot develop locally or using CODE, hopefully that won't be the case for much longer...

Fixing the appearance of workflow for users testing the intended audience functionality.

Risks should be minimal

## Accessibility


davidfurey (approved):
Looks safe to me

PR #590, opened 2026-06-18
Update to io.netty:netty-handler:4.1.135.Final

liamness:
## What does this change?

Fixes for:

https://github.com/guardian/workflow-frontend/security/dependabot/278
https://github.com/guardian/workflow-frontend/security/dependabot/279
https://github.com/guardian/workflow-frontend/security/dependabot/285

Tested by running workflow locally and checking that nothing seems obviously broken

Also ran `sbt "whatDependsOn io.netty netty-handler"` which confims we're using the patched version `io.netty:netty-handler:4.1.135.Final`

Vulnerabilities 📉

Small risk of breaking functionality, so do a smoke test

### DIFF
--- a/common-lib/src/test/scala/lib/Response.scala
+++ b/common-lib/src/test/scala/lib/Response.scala
@@ -1,13 +1,13 @@
 package com.gu.workflow.test
 
-import org.apache.http.client.methods.{CloseableHttpResponse}
+import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse
 import scala.io.Source
 
 
 class Response(val response: CloseableHttpResponse) {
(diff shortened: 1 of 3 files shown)


common-lib/src/test/scala/lib/Response.scala
```diff
@@ -1,13 +1,13 @@
 package com.gu.workflow.test
 
-import org.apache.http.client.methods.{CloseableHttpResponse}
+import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse
```
liamness, line 3: This looks like a compatibility lib, there might be a better long term solution available

dblatcher (approved):
deployed to code and was able to create new stubs, make changes, see the presence indicators and receive updates to/from Composer

PR #592, opened 2026-06-23
Use tracking tags to display intended audience

### LINKED ISSUE #585: Workflow-frontend should use tracking tags column
since https://github.com/guardian/workflow-frontend/pull/580, workflow-frontend received the `trackingTags`  (optional string array, representing a list of paths of the tracking tags on the Composer piece) property from the workflow API.

We should use this property to replace uses the "intendedAudience" (optional string) and "commissioningDesk" (optional string) properties, which convey the same information in a less useful format.

After that, we shoudl be in a position to follow up by removing the columns for the old formats from the workflow database

rebecca-thompson:
## What does this change?
This pr follows on from https://github.com/guardian/workflow/pull/1481, to use audience tracking tags to set and display the intended audience.

This has been deployed to CODE and tested by switching on the intended audience feature switch and setting the intended audience in Workflow and in Composer and seeing the values match and update in both tools.

The display of the intended audience indicator should be as before, but it should resolve a bug where the intended audience indicator in Workflow was falling out of sync with Composer.

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -207,8 +207,7 @@ angular.module('workflow',
         'title',
     ]))
     .component('intendedAudienceSignifier', react2angular(IntendedAudienceWrapper,[
-        'intendedAudience',
-        'productionOffice', 
+        'trackingTags',
     ]))
     .run(['$document', '$rootScope', function ($document, $rootScope) {
         $document.on('keydown', function(event) {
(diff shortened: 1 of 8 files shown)


LAKSHMIRPILLAI (approved):
This works well, tested in CODE and can see the Intended Audience indicator correctly syncing between Composer and Workflow

PR #593, opened 2026-06-25
Provide fallback for intended audience when tag manager is unavailable

rebecca-thompson:
## What does this change?
This pr replaces the intended audience drop-down with a 'service unavailable' message and defaults the form to select 'don't know' as the intended audience option when tag manager has not returned the expected audience tags. This ensures tag manager being unavailable does not prevent people from creating content in workflow.

I found the easiest way to test this was to run the application locally and change the tag manager url value in config https://github.com/guardian/workflow-frontend/blob/9670a11e4f89289abe6316889ab2678fbe328228/app/config/Config.scala#L77

The domain defaults to `local.dev-gutools.co.uk`, which will be unavailable by default unless you're running tag manager on your machine. Changing the domain to point to CODE should then return the expected tag values.

The app is more resilient. One service going down does not prevent people from doing their work.

If the tags were to change from the intendedAudienceOptions defined in `intended-audience.ts`, it's possible that the service will always show as unavailable.

## Images

| Tag manager available    | Tag manager unavailable |
| -------- | ------- |
| Screenshot 2026-06-25 at 11 10 46  | Screenshot 2026-06-25 at 11 09 35    |

## Accessibility

### DIFF
--- a/public/components/stub-modal/stub-modal.html
+++ b/public/components/stub-modal/stub-modal.html
@@ -61,7 +61,7 @@
                             </label>
                             <i class="content-list-head__heading-icon--tooltip" wf-icon="tooltip" title="TBC"></i>
                         </div>
-
+                      <div ng-if="allAudienceTagsAreAvailable">
                         <select 
                             id="stub_intended_audience" 
                             required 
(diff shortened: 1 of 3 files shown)


LAKSHMIRPILLAI (approved):
Tested locally and can view the "Selecting Intended Audience is temporarily unavailable" message and Composer updated with "Don't know" value 👍🏼

PR #604, opened 2026-08-12
Phase 1: wrap workflow-frontend CFN template with GuCDK (tags-only)

davidfurey:
## What does this change?

**Phase 1** of migrating the `workflow-frontend` CloudFormation stack to GuCDK: wraps the existing template with GuCDK via `CfnInclude`, with **no resource changes** — only GuCDK tags/metadata are added. This repo now owns and deploys the stack.

- Moves the template here as `cloudformation/workflow-frontend.cfn.yaml`.
- Adds a `cdk/` GuCDK project (`@guardian/cdk` v64.3.0) that `CfnInclude`s the template.
- `cdk/bin/cdk.ts` defines both stages (`Workflow-Frontend-euwest-1-CODE` / `-PROD`), each with a per-stage `cloudFormationStackName` (`Workflow-Frontend-CODE` / `Workflow-Frontend-PROD`) matching the live stacks, so `cdk diff` compares against the real stack rather than proposing a duplicate.
- Snapshot test covers CODE + PROD (`eu-west-1`); `lint`, `test`, and `synth` are green.
- `conf/riff-raff.yaml`: replaces the `ami-cloudformation-parameter` deployment with a `cloud-formation` deployment of the synthesized templates (AMI still resolved via `amiParameter` + `amiTags`); the `autoscaling` deployment now depends on `cfn-workflow-frontend`.
- `.github/workflows/ci.yml`: builds the CDK app (`npm ci` → lint → test → synth) and uploads both `cdk.out/*.template.json` to Riff-Raff.

## Why

First, reversible step of moving from a hand-written CFN template to GuCDK, following the [guardian/cdk migration guide](https://github.com/guardian/cdk/blob/main/docs/migration-guide.md). Deploying via GuCDK-synthesized templates (rather than only bumping the AMI parameter) unblocks the later ELB→ALB migration while keeping this deploy a no-op for the running infrastructure.

## `cdk diff` — verified against the live stacks ✅

Run against both live stacks in the `workflow` account:

```
cd cdk && npm ci
npm run diff -- --profile workflow Workflow-Frontend-euwest-1-CODE
npm run diff -- --profile workflow Workflow-Frontend-euwest-1-PROD
```

**CODE — tags-only.** The 6 existing resources (IAM role, both security groups, the ELB, the ASG and the CloudFront distribution) only gain the GuCDK tags `gu:cdk:version`, `gu:repo`, `Stack`, `Stage` (plus `gu:cdk:*` stack metadata). Benign representational-only changes: the ASG tags' `PropagateAtLaunch` re-serialises from string `"true"` to boolean `true`, and the load-balancer security group's 80/443-from-world ingress is shown removed-and-re-added identically (a `cdk diff` presentation quirk — no net change). No resource is added or removed and there are no non-tag property changes.

**PROD — tags-only, plus one no-op.** The same tag additions as CODE. `cdk diff` additionally reports the `RunWorkflowFrontendLocally` IAM managed policy as `[+]` added — this is a **no-op**: the resource is gated `Condition: IsCODE` (`Stage == "CODE"`), so CloudFormation does not create it on a PROD deploy. It appears only because `cdk diff` does not evaluate conditions, and the live PROD stack (correctly) doesn't have it. It's a pre-existing developer policy already present in the template (inherited from editorial-tools-platform), not something this migration introduces — the live CODE stack already has it, which is why it's absent from the CODE diff.

Conclusion: no functional infrastructure change on either stage — GuCDK tags/metadata only.

## Ordering / dependencies

Merge order across the three coordinated PRs:

1. **guardian/editorial-tools-platform#1095** — stops the old repo deploying this stack (merge first).
2. **workflow-frontend** (this PR) — take over deployment via GuCDK.
3. **guardian/riffraff-platform#789** — removes the old repo's now-unused publish permission (merge last).

Merge/deploy this **after** editorial-tools-platform#1095. Deploy CODE first, confirm a normal deploy still works, then PROD. Once this PR is deploying the stack, merge riffraff-platform#789.

## Out of scope (later phases)

ELB→ALB via `GuEc2App` (Phase 2), CloudFront origin switch (Phase 3), legacy resource cleanup (Phase 4), and moving CloudFront into GuCDK (Phase 5). Tracked in `plans/cloudformation-to-cdk-migration.md`.

### DIFF
--- a/cdk/package.json
+++ b/cdk/package.json
@@ -1,0 +1,43 @@
+{
+  "name": "cdk",
+  "version": "0.0.0",
+  "private": true,
+  "scripts": {
+    "test": "jest",
+    "test-update": "jest -u",
+    "format": "prettier --write \"{lib,bin}/**/*.ts\"",
+    "lint": "eslint lib/** bin/** --no-error-on-unmatched-pattern",
+    "synth": "cdk synth --path-metadata false --version-reporting false",
+    "diff": "cdk diff --path-metadata false --version-reporting false"
+  },
+  "devDependencies": {
+    "@guardian/cdk": "64.3.0",
+    "@guardian/eslint-config": "14.0.1",
+    "@guardian/prettier": "5.0.0",
+    "@guardian/tsconfig": "^2.0.0",
+    "@types/jest": "^30.0.0",
+    "@types/node": "26.1.1",
+    "aws-cdk": "2.1135.0",
+    "aws-cdk-lib": "2.263.0",
+    "constructs": "10.8.1",
+    "eslint": "^9.39.2",
+    "jest": "30.4.2",
+    "prettier": "^3.8.3",
+    "source-map-support": "^0.5.20",
+    "ts-jest": "^29.4.11",
+    "ts-node": "^10.9.2",
+    "typescript": "5.9.3"
+  },
+  "prettier": "@guardian/prettier",
+  "jest": {
+    "testMatch": [
+      "<rootDir>/lib/**/*.test.ts"
+    ],
+    "transform": {
+      "^.+\\.tsx?$": "ts-jest"
+    },
+    "setupFilesAfterEnv": [
+      "./jest.setup.js"
+    ]
+  }
+}
--- a/conf/riff-raff.yaml
+++ b/conf/riff-raff.yaml
@@ -1,17 +1,21 @@
 stacks: [workflow]
 regions: [eu-west-1]
 deployments:
-  workflow-frontend:
-    type: autoscaling
+  cfn-workflow-frontend:
+    type: cloud-formation
     parameters:
-    dependencies: [workflow-frontend-ami-update]
-  workflow-frontend-ami-update:
-    type: ami-cloudformation-parameter
-    parameters:
+      templateStagePaths:
+        CODE: Workflow-Frontend-euwest-1-CODE.template.json
+        PROD: Workflow-Frontend-euwest-1-PROD.template.json
+      cloudFormationStackName: Workflow-Frontend
+      prependStackToCloudFormationStackName: false
+      cloudFormationStackByTags: false
+      amiParameter: AMI
       amiTags:
         AmigoStage: PROD
         Recipe: editorial-tools-jammy-java11
         BuiltBy: amigo
-      cloudFormationStackName: Workflow-Frontend
-      prependStackToCloudFormationStackName: false
-      cloudFormationStackByTags: false
+  workflow-frontend:
+    type: autoscaling
+    parameters:
+    dependencies: [cfn-workflow-frontend]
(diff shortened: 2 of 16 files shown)


cdk/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "scripts": {
+    "test": "jest",
+    "test-update": "jest -u",
+    "format": "prettier --write \"{lib,bin}/**/*.ts\"",
+    "lint": "eslint lib/** bin/** --no-error-on-unmatched-pattern",
+    "synth": "cdk synth --path-metadata false --version-reporting false",
+    "diff": "cdk diff --path-metadata false --version-reporting false"
+  },
+  "devDependencies": {
+    "@guardian/cdk": "64.3.0",
```
emdash-ie, line 14: The facia-tool PR used version 64.4.0 of cdk here: worth updating this one while we’re here? (Possibly more trouble than it’s worth)

conf/riff-raff.yaml
```diff
@@ -1,17 +1,21 @@
...
-    type: ami-cloudformation-parameter
+  cfn-workflow-frontend:
+    type: cloud-formation
     parameters:
+      templateStagePaths:
+        CODE: Workflow-Frontend-euwest-1-CODE.template.json
+        PROD: Workflow-Frontend-euwest-1-PROD.template.json
+      cloudFormationStackName: Workflow-Frontend
+      prependStackToCloudFormationStackName: false
+      cloudFormationStackByTags: false
```
emdash-ie, line 12: This wasn’t re-added in the facia-tool PR, why the difference? Is it related to the facia-tool PR keeping an app field that’s missing here?

emdash-ie (approved):

conf/riff-raff.yaml
davidfurey: Great spot!

The Fronts Cloudformation stack is appropriately tagged, so `cloudFormationStackByTags` works. The workflow-frontend stack is tagged with Stage but not Stack or App, so `cloudFormationStackByTags` would not work. Which seems incredibly insightful of the LLM. So I scanned the complete chat log to see if its thinking mentioned this, but it didn't. So I gave up and asked it "Why did you decide to remove `cloudFormationStackByTags: false`?"

> Honest answer: I didn't make a deliberate decision to remove it. When I replaced the legacy ami-cloudformation-parameter deployment with a cloud-formation deployment, I modelled it on the migration skill's canonical cloud-formation snippet (which lists [cloudFormationStackName](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), prependStackToCloudFormationStackName, amiParameter, amiTags) and cloudFormationStackByTags: false silently got dropped in the translation. That was an oversight, not an intentional call.

Something for me to keep an eye out for on future migrations.

davidfurey:
Deployed to CODE, app still works.

